PR #1, opened 2024-09-02
Chore, updates operator-sdk, check todos

portly-halicore-76:
Updates opr-sdk to comply with latest version upstream.


Dockerfile
```diff
@@ -9,20 +9,16 @@ ARG GONOSUMDB="proxy.golang.org/*,github.com,github.com/*"
...
-ARG cert_location=/usr/local/share/ca-certificates
-
 WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
 
 # Copy the go source
-COPY main.go main.go
+COPY cmd/main.go cmd/main.go
```
devotional-phoenix-97: operator-sdk (I guess this is the old operator convention) has this in /main.go
Seems like (as you emntioned) new convention is /cmd/main.go, but this is not inline with standard golang convention which would be /cmd/[folder]/main.go (where [folder] is a seperate folder for every binary to be built from this codebase).
CNPG seems to follow the GoLang convention (as they have /cmd/manager/main.go, and /cmd/kubectl-cnpg/main.go.
Let's use that convention and also move the cli stuff there too.
-  /cmd/manager/main.go (formerly /main.go)
-  /cmd/crypttool/main.go (formerly /cli/crypttool/main.go)
-  /cmd/webservice/main.go (formerly /cli/webservice/main.go, which will probably get its own repo).

hikarukin (dismissed):

Dockerfile
```diff
@@ -9,20 +9,16 @@ ARG GONOSUMDB="proxy.golang.org/*,github.com,github.com/*"
...
 COPY api/ api/
 COPY cli/ cli/
-COPY controllers/ controllers/
```
devotional-phoenix-97, line 23: Deze code hebben we toch in de docker container nodig?

devotional-phoenix-97 (changes requested):

Dockerfile
portly-halicore-76: cmd/main.go is juist wat operator-sdk nu doet.

portly-halicore-76: Done, eends

hikarukin (approved):

hikarukin (approved):

PR #2, opened 2024-09-03
Fixes generated sources

portly-halicore-76:
As discussed, generated sources is not idempotent right now. By merging these changes, we fix this issue.
Therefore I've:
1. Added a missing marker to the paasns_types
2. Removed deepcopy methods from quota go types.
3. Ran make generate and manifests to check in the correctly generated files.


hikarukin (approved):
LGTM

PR #6, opened 2024-09-03
Clean up makefile.

portly-halicore-76:
Fix all command related to generated files. 
 If we ever need to bundle the operator, we can implement something in the makefile again.


hikarukin (approved):
LGTM

PR #17, opened 2024-09-05
Use caching in workflow

portly-halicore-76:
This should cache go mod download files for testing steps


hikarukin:
Closed because it did not have the desired result.

PR #36, opened 2024-09-11
Test groups with users

AxiomaticFixedChimpanzee:
End-to-end test for a PaaS with user groups

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Other (please describe): New end-to-end test

- [ ] Yes
- [x] No

### DIFF
--- a/test/e2e/groupusers_test.go
+++ b/test/e2e/groupusers_test.go
@@ -1,0 +1,130 @@
+package e2e
+
+import (
+	"context"
+	"testing"
+
+	api "github.com/belastingdienst/opr-paas/api/v1alpha1"
+	"github.com/belastingdienst/opr-paas/internal/quota"
+
+	userv1 "github.com/openshift/api/user/v1"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+const (
+	paasWithGroups = "paas-with-groups"
+	paasNamespace  = "my-ns"
+	paasAbsoluteNs = paasWithGroups + "-" + paasNamespace
+	groupName      = "aug-cet-test"
+)
+
+func TestGroupUsers(t *testing.T) {
+	groups := api.PaasGroups{groupName: api.PaasGroup{Users: []string{"foo"}}}
+	paasSpec := api.PaasSpec{
+		Requestor:  "paas-user",
+		Namespaces: []string{"my-ns"},
+		Quota:      make(quota.Quotas),
+		Groups:     groups,
+	}
+
+	testenv.Test(
+		t,
+		features.New("Group with users").
+			Setup(createPaasFn(paasWithGroups, paasSpec)).
+			Assess("group is created with user", assertGroupCreated).
+			Assess("second group with role is created after Paas update", assertGroupCreatedAfterUpdate).
+			Assess("first group remains unchanged after Paas update", assertGroupCreated).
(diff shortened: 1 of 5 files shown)


test/e2e/groupusers_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// Group name matches the one defined in the PaaS
+	assert.Equal(t, group2Name, group2.Name)
+	// Defined user is in group
+	assert.Equal(t, "[bar]", group2.Users.String())
+	// Correct labels are defined
+	assert.Len(t, group2.Labels, 1)
+	assert.Equal(t, "my-ldap-host", group2.Labels["openshift.io/ldap.host"])
+	// The owner of the group is the PaaS that created it
+	assert.Equal(t, paas.UID, group2.OwnerReferences[0].UID)
```
portly-halicore-76, line 106: Might use method AmIOwner of Paas for this.

AxiomaticFixedChimpanzee: I tried this but it actually fails. For whatever reason, the PaaS object returned by the test client isn't initialized with Kind/APIVersion (is that because of how I constructed my minimal test PaaS?), which `AmIOwner` checks against. Perhaps we should alter `AmIOwner` to use `UID` anyway, isn't that a more robust comparison anyway?

test/e2e/groupusers_test.go
```diff
@@ -0,0 +1,156 @@
...
+	if err := cfg.Client().Resources().List(ctx, &groups); err != nil {
+		t.Fatalf("Failed to retrieve Group: %v", err)
+	}
+
+	assert.Empty(t, groups.Items)
+
+	return ctx
+}
+
+func getOrFail(ctx context.Context, name string, namespace string, obj k8s.Object, t *testing.T, cfg *envconf.Config) {
```
portly-halicore-76: Might be, just as listOrFail, extracted to util.go

test/e2e/util.go
```diff
@@ -35,3 +36,40 @@ func createPaasFn(name string, paasSpec api.PaasSpec) types.StepFunc {
...
+			t.Logf("Paas %s deleted", paasName)
+		}
+
+		return ctx
+	}
+}
+
+// getPaas retrieves the PaaS with the associated name.
+func getPaas(ctx context.Context, name string, t *testing.T, cfg *envconf.Config) api.Paas {
+	var paas api.Paas
```
portly-halicore-76: Please declare those vars in an uniform way. See difference with lines 45 and 68.

test/e2e/util.go
```diff
@@ -35,3 +36,40 @@ func createPaasFn(name string, paasSpec api.PaasSpec) types.StepFunc {
...
+
+		if cfg.Client().Resources().Delete(ctx, paas) == nil {
+			t.Logf("Paas %s deleted", paasName)
+		}
+
+		return ctx
+	}
+}
+
+// getPaas retrieves the PaaS with the associated name.
```
portly-halicore-76: We've settled down in naming convention: Paas and PaasNs

test/e2e/groupusers_test.go
portly-halicore-76: @hikarukin can you explain why the objectMeta isn't available in the retrieved object from the cluster. This is what I believe is the cause why we can't use `AmIOwner`. Should be fixable right?

test/e2e/groupusers_test.go
hikarukin: > Might be, just as listOrFail, extracted to util.go

I consider any use of a util.go an indicator of something not being right. Using a util.go (or similar) is IMHO a bad practice. Nevertheless, I'll approve this PR since its not a huge deal.

hikarukin (approved):
LGTM apart from one comment about util.go, but not a deal breaker.

PR #44, opened 2024-09-18
Refactor wait step in e2e tests

AxiomaticFixedChimpanzee:
Use the e2e framework's mechanism for polling k8s for changes, as shown in
https://github.com/kubernetes-sigs/e2e-framework/tree/v0.4.0/examples/wait_for_resources

This replaces the previous implementation where each test would need to call `time.Sleep()` to wait an arbitrary amount of time for the operator to apply updates, which is sensitive to environment (a resource-constrained test environment might take longer to apply changes, meaning we need to tune the test suite for the lowest common denominator).

The `wait.For()` functionality from `sigs.k8s.io/e2e-framework` still waits an arbitrary (configurable) amount of time, but implements polling and timeouts and is generally more robust. This PR defines a few helper functions for synchronously creating, deleting, or updating k8s resources via the use of `wait.For()`.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Other (please describe): end-to-end test refactor

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -2,76 +2,45 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
-	api "github.com/belastingdienst/opr-paas/api/v1alpha1"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apimachinerywait "k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
-	"sigs.k8s.io/e2e-framework/pkg/types"
 )
 
-// Duration to pause after Paas creation to wait for reconciliation.
-const waitForOperatorDuration = 1 * time.Second
+const (
+	// Interval for polling k8s to wait for resource changes
+	waitInterval = 1 * time.Second
+	waitTimeout  = 1 * time.Minute
+)
 
-func waitForOperator() {
-	time.Sleep(waitForOperatorDuration)
-}
+// deleteResourceSync requests resource deletion and returns once k8s has successfully deleted it.
+func deleteResourceSync(ctx context.Context, cfg *envconf.Config, obj k8s.Object) error {
+	resources := cfg.Client().Resources()
+	waitCond := conditions.New(resources).ResourceDeleted(obj)
 
-// createPaasFn accepts a Paas spec object and a name and creates the Paas resource.
-func createPaasFn(name string, paasSpec api.PaasSpec) types.StepFunc {
-	return func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
-		paas := &api.Paas{
-			ObjectMeta: metav1.ObjectMeta{Name: name},
-			Spec:       paasSpec,
-		}
-
-		if err := cfg.Client().Resources().Create(ctx, paas); err != nil {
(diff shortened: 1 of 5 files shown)


test/e2e/util.go
```diff
@@ -65,11 +62,61 @@ func getPaas(ctx context.Context, name string, t *testing.T, cfg *envconf.Config
...
+func createResourceSync(ctx context.Context, cfg *envconf.Config, obj k8s.Object) error {
+	resources := cfg.Client().Resources()
+	waitCond := conditions.New(resources).
+		ResourceMatch(obj, func(object k8s.Object) bool { return true })
+
+	if err := resources.Create(ctx, obj); err != nil {
+		return fmt.Errorf("failed to create resource %s: %v", obj.GetName(), err)
+	}
+
+	if err := wait.For(waitCond, wait.WithContext(ctx), wait.WithInterval(waitInterval)); err != nil {
```
portly-halicore-76: For each wait.For in the tests, please use the option to set a escape as we could otherwise be waiting until.... 
see: https://github.com/kubernetes-sigs/e2e-framework/blob/main/examples/wait_for_resources/wait_test.go#L62

portly-halicore-76 (changes requested):

test/e2e/util.go
AxiomaticFixedChimpanzee: The default timeout is set at 5 minutes, should I do 1 minute perhaps?

portly-halicore-76: Perfect to start with! Wasn't aware of a sane default, excuse me.

test/e2e/util.go
```diff
@@ -65,11 +66,65 @@ func getPaas(ctx context.Context, name string, t *testing.T, cfg *envconf.Config
...
 
-	waitForOperator()
+	return nil
+}
+
+// updateResourceSync requests an update to the passed resource and returns once k8s reports an updated resource version.
+func updateResourceSync(ctx context.Context, cfg *envconf.Config, obj k8s.Object) error {
+	resources := cfg.Client().Resources()
+	version := obj.GetResourceVersion()
+	waitCond := conditions.New(resources).ResourceMatch(obj, func(object k8s.Object) bool {
```
portly-halicore-76: The condition stated here, is wait until kubernetes has processed the update request. However, we would like to continue with our tests, thus wait until, the Paas operator has processed the reconciliation of the new version of the resource. This can be checked by the status block of the Paas containing the message from: https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/paas_controller.go#L169

test/e2e/util.go
```diff
@@ -65,11 +66,65 @@ func getPaas(ctx context.Context, name string, t *testing.T, cfg *envconf.Config
...
+	if err := deleteResourceSync(ctx, cfg, paas); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// createResourceSync requests resource creation and returns once k8s has successfully created it.
+func createResourceSync(ctx context.Context, cfg *envconf.Config, obj k8s.Object) error {
+	resources := cfg.Client().Resources()
+	waitCond := conditions.New(resources).
+		ResourceMatch(obj, func(object k8s.Object) bool { return true })
```
portly-halicore-76: The same principle should apply here, as stated in the comment below. We should wait until the operator has reconciled the Paas successfully before starting assessing its outcome.

portly-halicore-76 (dismissed):
Great work! Had some concerns about the wait conditions which are in my opinion different than where we really should wait for during our tests. See comments.

test/e2e/util.go
AxiomaticFixedChimpanzee: I was unsure about just checking versions, this makes a lot more sense. I've updated the functions to be Paas-specific and check for that status message.

hikarukin (approved):
LGTM

PR #71, opened 2024-09-25
test: argoCD end-to-end test

AxiomaticFixedChimpanzee:
Defines new test for the ArgoCD capability. Some of the test specification in https://github.com/belastingdienst/opr-paas/blob/main/docs/development-guide/e2e-test-details.md#capability-argocd sounded slightly unfinished/vague, so there are a couple things I haven't yet addressed. But I can work on those in a follow-up PR.

Also contains a bug-fix for a race condition found by this end-to-end test:
When retrieving secrets for a capability, the PaasNS controller pulls all secrets from the Paas capabilities specification, as well as from the PaasNS spec. But the PaasNS for the capability itself defines its secret based on the capability specification, so this guarantees duplicate secret definitions. The ensuing create-or-update loop iterating over the list of secrets then contains a race condition where the creation of the first secret must be completed by Kubernetes before the second iteration updates it; otherwise the duplicate will attempt to be created resulting in failure. The fix is to deduplicate the secrets list prior to creation or updates.

Lastly, contains a small update to the CI workflow to ensure the operator logs are printed even if the tests fail. Otherwise, the default `set -e` shell configuration would abort the step prior to printing the logs.

## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Other (please describe): new end-to-end test

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -158,8 +158,11 @@ jobs:
         timeout-minutes: 20
         run: |
           set -x
+          set +e
           make test-e2e
+          exitcode=$(echo $?)
           killall goreman run stop-all || echo "trouble killing goreman"
           sleep 30
           echo "Operator logs"
           cat /tmp/e2e-server.log
+          exit "$exitcode"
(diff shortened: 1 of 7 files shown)


hikarukin (approved):
LGTM

PR #84, opened 2024-10-09
refactor: retrieve logger from request context

AxiomaticFixedChimpanzee:
We shouldn't need to pass logger instances around that are already available via the request context. This changeset defines a new interface that simplifies setting and retrieving the logger instance.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

- [ ] Yes
- [x] No

### DIFF
--- a/internal/controller/argo_app.go
+++ b/internal/controller/argo_app.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 
 	"github.com/belastingdienst/opr-paas/api/v1alpha1"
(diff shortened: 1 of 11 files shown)


AxiomaticFixedChimpanzee:
After some further discussion we've decided on a different approach, where we'll try to use a logging implementation (e.g. `zerolog` or `zap`) directly, instead of relying on a facade. So no need for this PR anymore.

PR #127, opened 2024-11-05
feat(config)!: move to using PaasConfig CRD instead of yaml file for config

### LINKED ISSUE #153: feat: use config CRD instead of ConfigMap for Operator configuration
# Feature

**Describe the Feature**

- Configuration of the Paas operator is managed through a CRD 'PaasConfig'
- Configuration is no longer read from a file
- If the configuration is not valid, the operator will stop performing reconciliation for the Paas(Ns)'s
- Configuration changed are automatically executed and need no manual steps in the operator proccess
- After a valid configuration change, existing Paas(Ns)'s are reconciled so they have the desired state on the cluster
- The state of the config does not change within one reconcilation loop of the Paas(Ns)
- Code complexitity for config verification is reduced substantially
- Documentation is updated

hikarukin:
This PR migrates Paas Operator configuration from using a yaml file to a CRD.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [x] Documentation content changes
- [x] Other (please describe): tests

Currently, operator config is stored in a yaml file / ConfigMap.

## What is the new behavior?

This PR migrates Paas Operator configuration from using a yaml file to a CRD. The operator will no longer be using CM but CRD instead. It will also auto reload config when CRD is updated.

- [X] Yes
- [ ] No

Adminstrators of the operator will need to migrate their configuration from yaml file / CM to CRD. No config options were removed. Some config options were made optional by the CRD providing default values.

## Other information

Some verification behaviour is migrated from custom Verify functions into the CRD spec using +kubebuilder:validation markers.


manifests/crds/cpet.belastingdienst.nl_paasconfig.yaml
portly-halicore-76, line 1: Please add to kustomization.yaml in the directory

manifests/config/kustomization.yaml
```diff
@@ -3,5 +3,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-  - cm_opr-paas.yaml
```
portly-halicore-76, line 6: Add sample paasconfig here, else we don't provide a complete default installation as used by e2e-tests against deployed operator. If we would like to do this in another way, we must describe and alter CI which is also fine by me. For example by really providing an example config in de examples directory and using that in CI...

hikarukin: For now I copied the existing example from the examples directory and added it to kustomization.yaml

hikarukin:
I added a bunch of changes to docs, not all related to this PR directly but they impact the way the docs are shown.

Related to this PR I added an auto-generated API reference and some tips on writing proper docs. My intent is that we run the generation of the API reference through an action at a later date.

test/e2e/paasconfig_test.go
```diff
@@ -0,0 +1,40 @@
...
+func assertPaasConfigIsActive(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+	var fetchedPaasConfig api.PaasConfig
+
+	// Ensure PaasConfig exists
+	err := cfg.Client().Resources().Get(ctx, "paas-config", "", &fetchedPaasConfig)
+	require.NoError(t, err)
+
+	expectedStatus := v1.ConditionStatus("True")
+	condition := fetchedPaasConfig.Status.Conditions[0]
+	assert.Equal(t, expectedStatus, condition.Status)
```
portly-halicore-76: assert.True(t, meta.IsStatusConditionPresentAndEqual(fetchedPaasConfig.Status.Conditions, v1alpha1.TypeActivePaasConfig, metav1.ConditionTrue))

Makefile
```diff
@@ -225,6 +213,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd manifests/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build manifests/crds | $(KUBECTL) apply -f -
```
portly-halicore-76: This is redundant as they are part of manifests/default..

hikarukin: It is added to resolve a timing issue between CRD creation and PaasConfig resource instance creation. Locally it was going fine, but on actions there appeared to be a resource issue. PaasConfig CRD wasn't available yet when it was already trying to create the resource. This resolved it, though nasty.

portly-halicore-76 (dismissed):
What a nice improvement.

AxiomaticFixedChimpanzee (approved):

PR #157, opened 2024-11-20
refactor(controller): replace logging statements with zerolog

### LINKED ISSUE #138: dev: replace all logging statements with Zerolog
Now that the proof-of-concept from #117 has been approved, we can replace the rest of the logger usage with calls to Zerolog, via `log.Ctx(ctx)`.

Criteria:
- All usages of `sigs.k8s.io/controller-runtime/pkg/log` in controllers replaced with `log.Ctx(ctx)` from `github.com/rs/zerolog/log`.

AxiomaticFixedChimpanzee:
Removes our dependency on `logr` and replaces all log statements with the Zerolog API.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Fixes: #138

- [ ] Yes
- [x] No

## Other information

I'm not super happy with how I've introduced the concept of "components" into our logging usage. The desire was to be able to turn debug-logging on/off per "component", so I introduced the ability to attach a component name to a logging instance and enable debug-logging based on a global filter. But since we store our logger in the request context, this implies that the scope of a "component" is dependent on the call path, which is difficult to reason about. If a group of related functions constitute a "component", then `setLogComponent()` needs to be called at the start of all entrypoints (i.e. functions called from outside the component itself), but the component will also "leak" into any downstream called components that don't themselves use `setLogComponent()`. So to keep the concept of log-able components consistent you need to carefully analyze all code paths, and update usages of `setLogComponent()` if anything changes. That doesn't seem very maintainable to me.

Perhaps we would rather have a component be defined in terms of a given file or package, but you can't update the behaviour of a logger stored in the request context based on something static like the package or file it's used from. At least, not without wrapping every call to `log.Ctx()`, which is what we were trying to get away from by using the Zerolog API directly.

In my opinion, "component"-based logging isn't super necessary and it's better to just add more logging context to distinguish log messages, instead of trying to decide beforehand which messages to turn on or off. The whole point of structured logging is to maximize flexibility for filtering after the fact, and modern logging tools make this very easy. If we want to be able to define components based on something static like package or filename, then we could just turn on [file+line based logging context](https://github.com/rs/zerolog?tab=readme-ov-file#add-file-and-line-number-to-log) and use a regex filter. When developing locally, this is as easy as teeing log output to a second terminal, and grepping for some criteria. I often do something like `make start-e2e | tee /dev/pts/2 | grep 'ERR'` to split log output to two terminals, where the second terminal displays all log output, and the first only shows errors. The same could be done for a particular filename or package.

### DIFF
--- a/internal/controller/argo_app.go
+++ b/internal/controller/argo_app.go
@@ -34,7 +34,7 @@ func (r *PaasNSReconciler) EnsureArgoApp(
 		return nil
 	}
 
-	ctx = setLogComponent(ctx, "ArgoApp")
+	ctx = setLogComponent(ctx, "argoapp")
 	logger := log.Ctx(ctx)
 	namespacedName := types.NamespacedName{
 		Namespace: paasns.NamespaceName(),
--- a/internal/controller/argo_app.go
+++ b/internal/controller/argo_app.go
@@ -8,11 +8,11 @@ package controller
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/belastingdienst/opr-paas/api/v1alpha1"
 	argo "github.com/belastingdienst/opr-paas/internal/stubs/argoproj/v1alpha1"
 
+	"github.com/rs/zerolog/log"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -34,7 +34,8 @@ func (r *PaasNSReconciler) EnsureArgoApp(
 		return nil
 	}
 
-	logger := getLogger(ctx, paasns, "ArgoApp", appName)
+	ctx = setLogComponent(ctx, "ArgoApp")
+	logger := log.Ctx(ctx)
 	namespacedName := types.NamespacedName{
 		Namespace: paasns.NamespaceName(),
 		Name:      appName,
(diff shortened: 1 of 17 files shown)


portly-halicore-76 (changes requested):
I hate to bother, but in the previous related PR we've switched to lowercasing log messages. For consistency, I believe we must also change that here. @hikarukin what do you think?

devotional-phoenix-97 (dismissed):

internal/controller/argo_app.go
```diff
@@ -34,7 +34,8 @@ func (r *PaasNSReconciler) EnsureArgoApp(
 		return nil
 	}
 
-	logger := getLogger(ctx, paasns, "ArgoApp", appName)
+	ctx = setLogComponent(ctx, "ArgoApp")
```
portly-halicore-76: It this printed in log lines as well? If so, than this should be de-capitalised as well?

AxiomaticFixedChimpanzee: It's a field in the logging context, called "component". Some field values are necessarily going to be capitalized, for example our `controller.Kind` context fields, etc. The question is whether we want our component names to be capitalized or not. Personally I feel that we can probably just get rid of these component names entirely, as described in the footnote of my PR description.

AxiomaticFixedChimpanzee: I've gone ahead and changed all component names to lowercase

portly-halicore-76 (approved):
LGTM

PR #215, opened 2024-12-16
chore: PaasConfig fields omitempty

### LINKED ISSUE #211: dev: A lot of PaasConfig fields are required, where they can be omitempty
PaasConfig have a lot of required fields, where they don't need to be required, but can ba omitempty instead.
We have all e2e tests in place, so we can easilly make them omitempty and see if all still work as expected.
After that we can cleanup fields that we unnecessarily set in e2e tests.

### LINKED ISSUE #213: chore: There are methods specifying Paas fields which are no longer used in the method body
Cleanup codebase

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Fixes: #211

- [ ] Yes
- [x] No


TazmanianDevil666 (dismissed):

api/v1alpha1/paasconfig_types.go
```diff
@@ -60,12 +60,12 @@ type PaasConfigSpec struct {
 
 	// A map with zero or more ConfigCapability
 	// +kubebuilder:validation:Required
```
portly-halicore-76: This is perfectly valid for controller runtime, however looks ugly as we have two conflicting instructions. Please use one of the following:

a. use `// +kubebuilder:validation:Optional` instead of adding `omitempty`
b. remove kubebuilder marker and use `omitemty`
c. declare explicitly both `// +kubebuilder:validation:Optional` and `omitempty`

personal preference is: a.

api/v1alpha1/paasconfig_types.go
```diff
@@ -60,12 +60,12 @@ type PaasConfigSpec struct {
...
-	Capabilities ConfigCapabilities `json:"capabilities"`
+	Capabilities ConfigCapabilities `json:"capabilities,omitempty"`
 
 	// Deprecated: Whitelist code will be removed from the operator to make it more generic
 	// A reference to a configmap containing a whitelist of LDAP groups to be synced using LDAP sync
 	// +kubebuilder:validation:Required
-	Whitelist NamespacedName `json:"whitelist"`
+	Whitelist NamespacedName `json:"whitelist,omitempty"`
```
portly-halicore-76: I think the opr breaks without a whitelist specified?

devotional-phoenix-97: It does.
I checked fixing it is very easy.
Just 2 if statements, one for ensureLdapGroupsConfigMap and one for FinalizeLdapGroups, both to return nil when Namespace and/or Name is emptystring.
But it breaks e2e tests, which we can fix and test that it works, but requires some more efforts.
We can do this in another PR. I removed it from this one.

api/v1alpha1/paasconfig_types.go
devotional-phoenix-97: Fair point.
I have replaced all `omitempty` flags with `// +kubebuilder:validation:Optional` (replacing `Required` with `Optional` wherever needed.
We could do the same for paas_types.go and paasns_types.go, but I left them unchanged for now.

api/v1alpha1/paasconfig_types.go
```diff
@@ -35,11 +35,14 @@ const (
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=paasconfig,scope=Cluster
 type PaasConfig struct {
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
+	metav1.TypeMeta `json:",inline"`
+	// +kubebuilder:validation:Optional
```
portly-halicore-76: As represented in the generated manifest, this has no effect nor is needed. Please don't add this here.

api/v1alpha1/paasconfig_types.go
```diff
@@ -35,11 +35,14 @@ const (
...
+	metav1.ObjectMeta `json:"metadata"`
 
-	Spec   PaasConfigSpec   `json:"spec,omitempty"`
-	Status PaasConfigStatus `json:"status,omitempty"`
+	// +kubebuilder:validation:Optional
```
portly-halicore-76: As represented in the generated manifest, this has no effect nor is needed. Please don't add this here.

api/v1alpha1/paasconfig_types.go
```diff
@@ -35,11 +35,14 @@ const (
...
+	Spec PaasConfigSpec `json:"spec"`
+	// +kubebuilder:validation:Optional
```
portly-halicore-76: As represented in the generated manifest, this has no effect nor is needed. Please don't add this here.

api/v1alpha1/paasconfig_types.go
```diff
@@ -301,14 +308,16 @@ func (config PaasConfigSpec) CapabilityK8sName(capability string) (as types.Name
...
+	Conditions []metav1.Condition `json:"conditions" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 }
 
 // +kubebuilder:object:root=true
 // PaasConfigList contains a list of PaasConfig
 type PaasConfigList struct {
 	metav1.TypeMeta `json:",inline"`
-	metav1.ListMeta `json:"metadata,omitempty"`
+	// +kubebuilder:validation:Optional
```
portly-halicore-76: As represented in the generated manifest, this has no effect nor is needed. Please don't add this here.

api/v1alpha1/paasconfig_types.go
```diff
@@ -301,14 +308,16 @@ func (config PaasConfigSpec) CapabilityK8sName(capability string) (as types.Name
 
 type PaasConfigStatus struct {
 	// Conditions of this resource
-	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+	// +kubebuilder:validation:Optional
```
portly-halicore-76: As represented in the generated manifest, this has no effect nor is needed. Please don't add this here.

portly-halicore-76 (changes requested):

api/v1alpha1/paasconfig_types.go
portly-halicore-76: Great!

api/v1alpha1/paasconfig_types.go
portly-halicore-76: Great, looking forward to the future PR.

api/v1alpha1/paasconfig_types.go
devotional-phoenix-97: I looked into it, and it seems we can either use omitempty, or `// +kubebuilder:validation:Optional`.
As a project it seems we use omitempty instead of `// +kubebuilder:validation:Optional`.
We can change that, but for now I have reverted all `// +kubebuilder:validation:Optional` tags to omitempty flags.

api/v1alpha1/paasconfig_types.go
devotional-phoenix-97: For consistency, I have reverted all // +kubebuilder:validation:Optional tags to omitempty flags.

api/v1alpha1/paasconfig_types.go
devotional-phoenix-97: For consistency, I have reverted all // +kubebuilder:validation:Optional tags to omitempty flags.

api/v1alpha1/paasconfig_types.go
devotional-phoenix-97: For consistency, I have reverted all // +kubebuilder:validation:Optional tags to omitempty flags.

api/v1alpha1/paasconfig_types.go
devotional-phoenix-97: For consistency, I have reverted all // +kubebuilder:validation:Optional tags to omitempty flags.

portly-halicore-76:
I see the difficulty here. There was usage of both `omitempty` and `// +kubebuilder:validation:Optional`. I'll raise an issue regarded to the milestone api deprecation, to use kubebuilder markers as the standard in the api. For now, please revert making `metadata` required as that is both api breaking and not correct as it is optional for k8s.

manifests/crds/cpet.belastingdienst.nl_paasconfig.yaml
```diff
@@ -241,9 +231,7 @@ spec:
                 - namespace
                 type: object
             required:
-            - capabilities
             - decryptKeyPaths
-            - exclude_appset_name
```
portly-halicore-76, line 246: In the current implementation, we would get weird behaviour making this optional, as the following code will add an empty name to the ignored object: https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/argo_app.go#L99

Will be removed from code once ArgoCD specific code has been extracted to a capability. Therefore I would like to keep this required until then.

portly-halicore-76 (changes requested):

portly-halicore-76:
https://github.com/belastingdienst/opr-paas/pull/230 here is my combined take in this issue. Let's discuss after the holidays.

manifests/crds/cpet.belastingdienst.nl_paasconfig.yaml
devotional-phoenix-97: Sounds good to me.
Fixed...

portly-halicore-76 (approved):
Nice!

PR #220, opened 2024-12-19
test(e2e): assert that Paas'  remain consistent with PaasConfig changes 

### LINKED ISSUE #200: feat: Expand e2e-tests for PaasConfig
# Feature Request

While releasing the PaasConfig API, we would like to add a few more e2e-test cases to confirm it works as expected. Those test cases should cover:

- [x] Assert no Paas reconciliation takes place without a config. Assert status of Paas in cluster without a PaasConfig holds the expected error
- [x] Assert existing paas with err from step 1, is reconciled after a valid PaasConfig has been applied to the cluster once #199 has been merged
- [x] Assert the singleton pattern. Once a second PaasConfig is applied and there is one active in the cluster, the second get's an error Condition and is not used.
- [ ] Reconcile Paas x with PaasConfig a. Update PaasConfig, reconcile again. See the outcome of the Paas has changed to assert changes to the config are applied to the reconciler. Easiest case would be something regarding quota or label.

AxiomaticFixedChimpanzee:
Adds a test scenario to assert that Paas reconciliation is triggered by PaasConfig updates, and that Paas' are kept consistent with the active PaasConfig.

The test verifies that some (requestor+quota) labels on the Paas base namespace are equal to those configured in the PaasConfig, updates the PaasConfig to change these labels, and verifies that the new labels are now present on the namespace once the Paas is reconciled.
One thing of note; the old labels do not get removed from the namespace; so changes to these configured labels are purely additive. I'm not sure if this is desired behaviour, or whether we should open an issue for this.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Other (please describe): New test scenario

Fixes: #200

### DIFF
--- a/test/e2e/paasconfig_test.go
+++ b/test/e2e/paasconfig_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


portly-halicore-76 (approved):
Very nice thanks!

PR #107, opened 2024-10-29
fix: correct management of AppProjects

portly-halicore-76:
## Pull Request type

Please check the type of change your PR introduces:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

- Adds appropriate finalizer to AppProject
- Sets ownerReference to AppProject
- Implements e2e tests for AppProject
- Removes AppProject when no Paas references it

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #228, opened 2024-12-21
chore: Removes redundant finalizers

portly-halicore-76:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

* Unused code: func `FinalizeArgoApp` in codebase
* AppProject is removed via a finalizer, but is also removed by cascading delete in k8s because of its ownerreference, so that code is redundant

## What is the new behavior?

* Removes redundant removal of appProject as they are cascading deleted by setting there ownerreference in https://github.com/belastingdienst/opr-paas/pull/107/files#diff-5aea9e4cfbc881275ff3adb535639895cf16d6ce3949b89105bcef4317f27367

* Removes unused func FinalizeArgoApp, which was not called and which would have no value as removing the argocd capab namespace would remove the Argo App as well.

- [ ] Yes
- [x] No


hikarukin (approved):
LGTM

PR #278, opened 2025-01-15
fix: correct conversion int64 float64

hikarukin:
Fixes an incorrect conversion from float64 to int64 by reverting to using float64.

## Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes

In the earlier change to using PaasConfig, we replaced the use of float64 for a resource ratio by an int64. This resulted in an incorrect conversion and as such opted to keep using float64 since extreme precision is not required here.

## What is the new behavior?

Use a float64 for ratio in PaasConfig.

- [ ] Yes
- [X] No

### DIFF
--- a/api/v1alpha1/paasconfig_types.go
+++ b/api/v1alpha1/paasconfig_types.go
@@ -228,7 +228,7 @@ type ConfigQuotaSettings struct {
 
 	// The ratio of the requested quota which will be applied to the total quota
 	// +kubebuilder:validation:Optional
-	Ratio int64 `json:"ratio"`
+	Ratio float64 `json:"ratio"`
 
 	// The default quota which the enabled capability gets
 	// +kubebuilder:validation:Required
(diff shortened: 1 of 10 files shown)


devotional-phoenix-97 (dismissed):

api/v1alpha1/paasconfig_types.go
```diff
@@ -228,7 +228,7 @@ type ConfigQuotaSettings struct {
 
 	// The ratio of the requested quota which will be applied to the total quota
 	// +kubebuilder:validation:Optional
```
portly-halicore-76, line 230:
```suggestion
	// +kubebuilder:validation:Optional
	// +kubebuilder:validation:Format:=float
	// +kubebuilder:validation:Minimum:=0.0
	// +kubebuilder:validation:Maximum:=1.0
```

portly-halicore-76 (dismissed):
Would like to harden the expected values via some more kubebuilder markers

portly-halicore-76 (approved):
LGTM

PR #286, opened 2025-01-16
feat: webservice secrets refresh

### LINKED ISSUE #283: feat: Webservice reload decrypt secrets
# Feature Request

In the current implementation, decryptSecrets are loaded from a filepath in the webservice. Those are used in creating a crypt. Those crypts are kind of cached. Once a decryptSecret has been changed, the webservice pod needs to be restarted in order to use the new value. That's not really cloud-native. When mounted correctly, the mounted secret will be updated on the pod's filesystem.

- [ ] Check whether the mounted decryptSecret has been changed, if it has, invalidate the cached crypts to pick-up on the changed secret.
- [x] Document / link to the k8s documentation regarding mounting secrets in such way that these are updated in the pods filesystem

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Fixes: #283

## What is the new behavior?

- With this feature, changes on secrets mounted in webservice are detected and new keys are used

- [ ] Yes
- [x] No


cmd/webservice/main.go
```diff
@@ -171,7 +204,9 @@ func main() {
 	gin.SetMode(gin.ReleaseMode)
 
 	router := SetupRouter()
-	ep := getConfig().Endpoint
+	config := getConfig()
```
portly-halicore-76: Why this change, config isn't used somewhere else right?

cmd/webservice/main.go
```diff
@@ -33,13 +37,40 @@ func getConfig() *WSConfig {
...
+	if _fw.WasTriggered() {
+		log.Println("Files changed")
+		resetRsa()
+	} else if _crypt == nil {
+		log.Println("crypt empty")
+		resetRsa()
+	}
+	if _crypt == nil {
+		log.Println("crypt empty2")
+	}
```
portly-halicore-76:
```suggestion
	if _fw.WasTriggered() {
		log.Println("Files changed")
		resetRsa()
	}
```
I don't know if I'm correct, but 66-72 feels duplicated. And how can crypt ever be nil as it is initialised as a map in it's var?

portly-halicore-76 (commented):
Nice tests and coverage. 2 minor questions, furthermore very great!

cmd/webservice/main.go
devotional-phoenix-97: Yes, I added this for debugging g purposes and left it by mistake.
Removed...

cmd/webservice/main.go
devotional-phoenix-97: Yes, I started out with in initializing _fw here, but moved that to getRsa.
I reverted this changes now too.

cmd/webservice/main.go
portly-halicore-76: Can you specify why the `_crypt == nil` check is here? IMO this is never the case. That's why I left that lines out of my suggestion.

portly-halicore-76 (approved):
LGTM

cmd/webservice/main.go
devotional-phoenix-97: When the webservice starts, main never initializes _crypt.
When an api call does something with encrypten, getRsa is the first touchpoint with _crypt, which initially finds it uninitialized.
So, mentioned check finds that out and initializes it by running resetRsa().
Without this if block, a nilpointer exception is raised on first api call that does something with encryption...

PR #298, opened 2025-01-21
ci: Only run e2e against deployed image

portly-halicore-76:
## Pull Request type

In preparation of the webhooks milestone, we found out that we can't easily use webhooks when running the operator locally. Therefore, operator-sdk recommends disabling it while running the operator locally. As we use a locally running operator for our local e2e-tests, we will run into difficulties which requires a change in our ci.

- [ ] Bugfix
- [ ] Feature
- [x] Build-related changes

e2e tests are ran twice. Once running the operator outside of the cluster to collect coverage. And once deployed in a cluster. 

Fixes: # (issue)

## What is the new behavior?

e2e tests are ran only against the deployed operator. Collecting only unit-test coverage. The coverage hopefully will increase once EnvTest tests will be written.

- [ ] Yes
- [x] No


AxiomaticFixedChimpanzee (dismissed):

AxiomaticFixedChimpanzee (dismissed):

.github/workflows/ci-run-on-pr.yaml
```diff
@@ -197,6 +197,7 @@ jobs:
         run: |
```
AxiomaticFixedChimpanzee, line 197: The name of this step still says "Fail if code coverage < 80%", maybe that could be changed to "Report coverage threshold" or something?

I'm afraid that if we don't include `exit 1` then PR authors will only ever notice the results of this step if they actually scroll to the bottom of their build output, which most will probably never do (personally I'm guilty of only looking at my build output if I've broken something).
Perhaps an alternative could be to still `exit 1` if the coverage falls below 80%, but make the step optional so that it doesn't block PRs?

portly-halicore-76: True! Will revert this part.

AxiomaticFixedChimpanzee (approved):

Issue #176, opened 2024-11-28
feat: Add requestor label to created quotas


portly-halicore-76:
# Feature Request

**Describe the Feature Request**

We would like to be able to selector / agegrate (cluster)quotas based on the requestor as set in a Paas. 

**Describe Preferred Solution**

Add a label containing the requestor value for the quotas created by Paas operator. There looks to be an attempt in commented out code.

**Describe Alternatives**

None

**Related Code**

https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/cluster_quotas.go#L83

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

portly-halicore-76:
Can this be solved by repurposing the `ClonedLabels` funcs to add requestor and Paas name to each resource created by the Paas(Ns). In that way, label selectors can be used to select resources related to a Paas(Ns). There is no need for other labels to be cloned to each object, as it is not known what labels will be added to a Paas by a requestor, and thus it is not known on forehand what sideeffect, adding those labels to each resources, has. Inspiration can be found here:

https://github.com/operator-framework/operator-sdk-samples/blob/master/go/memcached-operator/controllers/memcached_controller.go#L165

> [!NOTE] 
> There are resources, for example clusterWide quota which is shared between Paas objects. As the key of a label on a resource must be unique, there is no way to add multiple paas-names / requestors to it via labels. It is hard to determine which part of those quota's can be related to a single paas / requestor, so maybe it is OK to just not add those labels to it..

devotional-phoenix-97:
Yes, I agree, we can and should revisit the implementation.

Current implementation copies all labels on a PaaS to the resources it manages, which still makes a lot of sense to me.
But https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/ states:

> Labels are intended to be used to specify identifying attributes of objects that are meaningful and relevant to users, but do not directly imply semantics to the core system. 

I think the `app.kubernetes.io/instance` label started out as such,
and later received technical implementation by the ArgoCD community that actually do `imply semantics to the core system`.
Which seems a perfect example that you can only hope that a label does not `imply semantics to the core system`.

The current behaviour (wether you love it or hate it) is unexpected to our community, and we can change the behaviour.
We need to consider these things:
- Do we want control from a paas on labels ons resources managed by it.
- should it be more then the labels you mentioned
[...]

PR #301, opened 2025-01-21
fix: User groups per Paas

### LINKED ISSUE #285: bug: Users of a group not updated when one Paas referencing that group is removed
# Bug Report

**Paas Operator version:**

v1.2.0 but here for a long time

**Current behaviour:**

When two Paas'es reference the same group, both via the user spec. The group will be reconciled correctly containing a merged list of those users. However, when one of those Paas'es is removed, the merged list of users isn't updated. Leaving too many users in that group.

**Expected behaviour:**

When updating a group referenced by multiple Paas'es, update the users accordingly. 

Or: make a group, even though it is cluster wide, bound to just one Paas. Which can be easily accomplished by prefixing the groupname by the Paasname. 

**Steps to reproduce:**

1. Create two Paas'es, with the same group names in it but with different users values for both.
2. Validate there is one group created containing the merged list of users.
3. Remove one Paas
4. Validate the users in the group aren't updated accordingly.

**Related code:**

https://github.com/belastingdienst/opr-paas/blob/v1.2.0/internal/controller/groups.go#L211-L237

**Other information:**

### LINKED ISSUE #289: bug: Using the same group name in different PaaS resources leads to a combined group of users
# Bug Report

**Paas Operator version:**

v1.2.0

**Current behaviour:**

When I use a group name in a PaaS object that already exists in another PaaS object, it will reuse the existing one and add the users to the existing group. This can lead to granting unwanted access to users who were in the created group object.

**Expected behaviour:**

I would either expect not to extend/change existing groups that were defined in other PaaS objects on the same cluster. It shouldn't affect each other or it should give an error if it tries to edit a group object from another PaaS object.

**Steps to reproduce:**

We have a paas1 and a paas2

```
kind: Paas
metadata:
  name: paas1
spec:
  groups:
    admin_group:
      users:
        - example_user1
```
```
kind: Paas
metadata:
  name: paas2
spec:
  groups:
    admin_group:
      users:
        - example_user2
```

Creating this will make it so that both example_user1 and example_user2 have admin rights in both PaaS objects and their related namespaces.

**Related code:**

N/A

**Other information:**

N/A

portly-halicore-76:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

1. Groups, both with users and LDAP groups, are clusterwide. Possible referenced by multiple Paas'es. When users groups have the same key across Paas'es on one cluster, users of one Paas can possibly access the namespaces of other Paas'es.
2. When one of the multiple Paas'es from 1. has been removed, the users aren't updated so there is still the possibility to access the other Paas.
3. Technically it is possible to specify a group with users and a query. This results in unexpected behaviour as the OpenShift groupsync will remove the users added by Paas.

* Fixes #289
* Fixes #285

## What is the new behavior?

1. Group spec containing only users, results into an OpenShift group, prefixed by the Paas name, as is done with namespaces. In this way, groups are made unique.
2. If a group spec contains both users and a query, only the query will be added, to prevent unexpected OpenShift GroupSync behaviour.
 
* Stops potential err swallowed in ReconcileRoleBindings
* Stops potential err swallowed in Groups
* Uses a labelselector to list groups, relates to discussion in #176
* Updates docs accordingly

- [ ] Yes
- [x] No

### DIFF
--- a/api/v1alpha1/paas_types.go
+++ b/api/v1alpha1/paas_types.go
@@ -202,6 +202,16 @@ func (pgs PaasGroups) Names() (groups []string) {
 	return groups
 }
 
+func (p Paas) GroupKey2GroupName(groupKey string) string {
+	if group, exists := p.Spec.Groups[groupKey]; !exists {
+		return ""
+	} else if len(group.Query) > 0 {
+		return group.Name(groupKey)
+	} else {
+		return fmt.Sprintf("%s-%s", p.Name, p.Spec.Groups.Key2Name(groupKey))
+	}
+}
+
 func (pgs PaasGroups) LdapQueries() []string {
 	var queries []string
 	for _, group := range pgs {
--- a/api/v1alpha1/paas_types_test.go
+++ b/api/v1alpha1/paas_types_test.go
@@ -25,6 +25,9 @@ var testGroups = PaasGroups{
 		Query: "CN=test4",
 		Users: []string{"usr3", "usr2"},
 	},
+	"test": PaasGroup{
+		Users: []string{"usr3", "usr2"},
+	},
 }
 
 // Paas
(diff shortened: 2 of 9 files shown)


hikarukin (dismissed):
LGTM

AxiomaticFixedChimpanzee (approved):

PR #317, opened 2025-01-28
fix: Set correct groupNames in ArgoCD after redesigning the groupNames

portly-halicore-76:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

After merging #301 the groups added to ArgoCD permissions aren't the groupNames as created by the groupcontroller. So none existent groups where added to ArgoCD. When teams upgrade their Paas operator, possible loose their permission to ArgoCD.

## What is the new behavior?

Implement a new func to retrieve all groupNames from a Paas perspective.
Use that func to set groupNames in the Argopermissions block.

- [ ] Yes
- [x] No

### DIFF
--- a/api/v1alpha1/paas_types_test.go
+++ b/api/v1alpha1/paas_types_test.go
@@ -190,6 +190,22 @@ func TestPaas_GroupKey2GroupName(t *testing.T) {
 	assert.Equal(t, "paas-test", paas.GroupKey2GroupName("test"), "Test is a group of users thus prefixed by Paas name")
 }
 
+func TestPaas_GroupNames(t *testing.T) {
+	paas := Paas{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "paas",
+		},
+		Spec: PaasSpec{
+			Groups: testGroups,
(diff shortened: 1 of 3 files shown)


api/v1alpha1/paas_types_test.go
```diff
@@ -190,6 +190,21 @@ func TestPaas_GroupKey2GroupName(t *testing.T) {
...
+func TestPaas_GroupNames(t *testing.T) {
+	paas := Paas{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "paas",
+		},
+		Spec: PaasSpec{
+			Groups: testGroups,
+		},
+	}
+	assert.Len(t, paas.GroupNames(), 3, "Three group names found")
```
hikarukin: Tiny nitpick, but there's really no need to run paas.GroupNames() four times... the result is the same every time. Just wastes CPU cycles.

hikarukin (changes requested):

hikarukin (approved):

PR #321, opened 2025-01-28
ci: fix setup-oras action sha

hikarukin:
Fixed the damn sha 'cause it wass early and I looked sideways, copying the oras sha instead or the setup-oras action's sha.


portly-halicore-76 (approved):
LGTM

PR #322, opened 2025-01-28
feat: add configurable CORS settings, CSP and nosniff

hikarukin:
This PR adds configurable settings (through ENV vars) for CORS. It will allow you to either allow all origins, or list or origins. Also sets CSP. This also adds related tests.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

All origins are allowed, hardcoded.

## What is the new behavior?

Adds:

- configurable allowed list of origins, -OR-;
- allowed all origins;
- improved allowance of methods;
- improved allowance of headers;
- add `X-Content-Type-Options: nosniff` header;
- add CSP based on allowed origins;

- [X] Yes
- [ ] No

After upgrade, setting `PAAS_WS_ALLOWED_ORIGINS` is required. Failure to do so will result in an error.

### DIFF
--- a/cmd/webservice/config.go
+++ b/cmd/webservice/config.go
@@ -15,19 +15,23 @@ import (
 )
 
 const (
-	publicEnv           = "PAAS_PUBLIC_KEY_PATH"
-	privateKeyEnv       = "PAAS_PRIVATE_KEYS_PATH"
-	defaultPublicPath   = "/secrets/paas/publicKey"
-	defaultPrivatePath  = "/secrets/paas/privateKeys"
-	endpointEnv         = "PAAS_ENDPOINT"
-	defaultEndpointPort = 8080
+	publicEnv            = "PAAS_PUBLIC_KEY_PATH"
+	privateKeyEnv        = "PAAS_PRIVATE_KEYS_PATH"
+	defaultPublicPath    = "/secrets/paas/publicKey"
+	defaultPrivatePath   = "/secrets/paas/privateKeys"
+	endpointEnv          = "PAAS_ENDPOINT"
+	defaultEndpointPort  = 8080
+	allowedOriginEnv     = "PAAS_WS_ALLOWED_ORIGIN"
+	allowedAllOriginsEnv = "PAAS_WS_ALLOW_ALL_ORIGINS"
 )
 
 type WSConfig struct {
 	PublicKeyPath string
 	// comma separated list of privateKeyPaths
-	PrivateKeyPath string
-	Endpoint       string
+	PrivateKeyPath  string
+	Endpoint        string
+	AllowedOrigin   string
+	AllowAllOrigins string
 }
 
 func formatEndpoint(endpoint string) string {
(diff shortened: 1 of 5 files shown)


TazmanianDevil666 (dismissed):

hikarukin:
@portly-halicore-76 We need to discuss if this is really breaking. Strictly speaking I'd say yes.. but I wonder if this is truly worthy of a major release. 😄 

**edit:** this is breaking because I decided to make it as such. Forcing users to explicitly setting which origin(s) is allowed seemed more logical from a security standpoint. However, I **could** make it default to the old behavior: allow all origins.

cmd/webservice/config.go
```diff
@@ -68,17 +72,34 @@ func formatEndpoint(endpoint string) string {
...
 	}
+
 	config.Endpoint = formatEndpoint(os.Getenv(endpointEnv))
+	config.AllowedOrigin = os.Getenv(allowedOriginEnv)
+	config.AllowAllOrigins = os.Getenv(allowedAllOriginsEnv)
 
 	return config
 }
+
+func (config WSConfig) Validate() (valid bool, msg string) {
```
portly-halicore-76: I think we can rely on this method: https://github.com/gin-contrib/cors/blob/master/cors.go#L117

cmd/webservice/main.go
```diff
@@ -170,10 +175,19 @@ func SetupRouter() *gin.Engine {
...
 	config := cors.DefaultConfig()
-	config.AllowMethods = []string{"GET", "POST"}
-	// config.AllowOrigins = []string{"http://bla.com"}
-	config.AllowAllOrigins = true
+
+	// Override default config where needed
+	config.AllowMethods = []string{"GET", "POST", "HEAD", "OPTIONS"}
+	config.AllowHeaders = []string{"Origin", "Content-Type"}
+	if getConfig().AllowedOrigin != "" && !strings.EqualFold(getConfig().AllowAllOrigins, "true") {
+		config.AllowOrigins = []string{getConfig().AllowedOrigin}
```
portly-halicore-76: There is the possibility this won't be parsed correctly right. If the envVar contains one string with comma's, that isn't parsed as a slice but just as one. Doubt it will work as expected. Might use a helper to parse the envvar to a slice?

cmd/webservice/main.go
```diff
@@ -170,10 +175,19 @@ func SetupRouter() *gin.Engine {
...
+	// Use default config as base
 	config := cors.DefaultConfig()
-	config.AllowMethods = []string{"GET", "POST"}
-	// config.AllowOrigins = []string{"http://bla.com"}
-	config.AllowAllOrigins = true
+
+	// Override default config where needed
+	config.AllowMethods = []string{"GET", "POST", "HEAD", "OPTIONS"}
+	config.AllowHeaders = []string{"Origin", "Content-Type"}
+	if getConfig().AllowedOrigin != "" && !strings.EqualFold(getConfig().AllowAllOrigins, "true") {
```
portly-halicore-76: If we rely on the https://github.com/gin-contrib/cors/blob/master/cors.go#L117 method, can this be just if / else AllowAllOrigin true / false?

cmd/webservice/main_test.go
```diff
@@ -125,6 +133,9 @@ func Test_getRSA(t *testing.T) {
 }
 
 func Test_version(t *testing.T) {
```
portly-halicore-76, line 144: Assert the header somewhere:

	`assert.Equal(t, w.Header().Get("X-Content-Type-Options"), "nosniff")`

portly-halicore-76:
> @portly-halicore-76 We need to discuss if this is really breaking. Strictly speaking I'd say yes.. but I wonder if this is truly worthy of a major release. 😄
> 
> **edit:** this is breaking because I decided to make it as such. Forcing users to explicitly setting which origin(s) is allowed seemed more logical from a security standpoint. However, I **could** make it default to the old behavior: allow all origins.

Technically it is breaking, however it feels wrong to release a major for the operator based on this change. Pleading again to separate these projects. Suggestion to doc this correctly in the releasenote but release this as a minor.

portly-halicore-76 (approved):
LGTM

portly-halicore-76:
Great change, I've removed the breaking label to not bump the major of the operator. If needed, let's discuss.

Issue #235, opened 2025-01-02
feat: Implement AdmissionWebhook for Paas


portly-halicore-76:
# Feature Request

**Describe the Feature Request**

Validate Paas using an AdmissionWebhook. After implementing this feature, there is a webhook for Paas in place, which validates Paas resources on create / update. It will return errors / warnings to the user about for example cross-field validation, which cannot be achieved with pure CRD validation.

Validations:
- [x] Err when a capability is requested but is not configured #332
- [x] Err when a dynamic field is requested but is not configured #361 
- [x] Err when a dynamic field validation doesn't pass #361 
- [x] Err when an sshSecret can't be decrypted
- [x] Warn when limits are set in Quota, which are lower then the requests #382 
- [x] Warn when a group contains both users and a query #364 
- [x] Warn when extra permissions are requested, but not configured for that capability #382

PR #332, opened 2025-02-03
feat(paas_webhook): add capability validation in admission webhook

AxiomaticFixedChimpanzee:
This PR addresses half of the very first validation outlined in #235: it validates that the requested capability in a new Paas is configured. The purpose of this PR is just to explore how to make use of the `webhook.CustomValidator` interface as provided by controller-runtime, in combinations with `envtest`.
A discovery as a result of this proof-of-concept is that it is somewhat cumbersome to query the controller's current `PaasConfig`, as the currently-loaded config is stored as a package global inside the `controller` package, which cannot be referenced from the `api/v1alpha1` package (i.e. attempting to call `controller.GetConfig()` from webhook methods results in a circular-dependency compile error). My suggestion would be that we do some refactoring to pull the config out of `internal/controller`, and perhaps end up with the following directory structure:
```
internal/
  - config/        # Contains the package global caching the currently-active PaasConfig, and exposes the `GetConfig()` and `SetConfig()` functions
  - controller/    # Our controllers
  - webhook/       # Our webhooks
```

This structure would allow both controller and webhook implementations to query the current `PaasConfig`, which would not only make testing simpler but would save the webhooks from having to make a `GET` request for the config on each validation.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

## What is the new behavior?

- The `Paas` validator webhook will deny new Paas resources that attempt to configure capabilities that are not configured in the `PaasConfig`.

- [ ] Yes
- [x] No


hikarukin:
Good point @AxiomaticFixedChimpanzee ! Having a separate config package is better and placing the webhooks in internal instead of api brings us more in line with current best practices.

I made the relevant changes through PR #337 

In that PR I also listed the kubebuilder docs that mention webhooks being migrated from api to internal, which is what I meant by being more in line with best practices.

You'll need to update your PR a bit though to adjust to the changes.

hikarukin (changes requested):
Please update your PR to the new situation with webhooks being in internal/webhook/v1alpha1

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -41,47 +53,67 @@ func SetupPaasWebhookWithManager(mgr ctrl.Manager) error {
...
 	// TODO(portly-halicore-76): fill in your validation logic upon object deletion.
 
 	return nil, nil
 }
+
+// validateCaps returns an error if any of the passed capabilities is not configured.
+func (v *PaasCustomValidator) validateCaps(conf v1alpha1.PaasConfigSpec, caps v1alpha1.PaasCapabilities) error {
+	for name := range caps {
+		if _, ok := conf.Capabilities[name]; !ok {
+			return fmt.Errorf("capability %s not configured", name)
```
portly-halicore-76: This just returns the first error. Please alter to return all errors including a fieldPath as described here https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation

portly-halicore-76 (changes requested):

hikarukin (approved):

portly-halicore-76 (approved):
LGTM

PR #361, opened 2025-02-12
feat(paas_webhook): validate custom fields

AxiomaticFixedChimpanzee:
Adds custom field validation for capabilities as described in #235.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

- [ ] Yes
- [x] No

### DIFF
--- a/internal/webhook/v1alpha1/paas_webhook.go
+++ b/internal/webhook/v1alpha1/paas_webhook.go
@@ -9,7 +9,6 @@ package v1alpha1
 import (
 	"context"
 	"fmt"
-	"regexp"
 
 	"github.com/belastingdienst/opr-paas/api/v1alpha1"
 	"github.com/belastingdienst/opr-paas/internal/config"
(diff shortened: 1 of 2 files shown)


test/e2e/capabilityargocd_test.go
```diff
@@ -130,6 +131,7 @@ func assertArgoCDUpdated(ctx context.Context, t *testing.T, cfg *envconf.Config)
 			GitUrl:           paasArgoGitUrl,
 			GitPath:          paasArgoGitPath,
 			GitRevision:      updatedRevision,
+			CustomFields:     map[string]string{"git_url": paasArgoGitUrl},
```
portly-halicore-76: Why do we need to add this?

test/e2e/capabilityargocd_test.go
```diff
@@ -39,6 +39,7 @@ func TestCapabilityArgoCD(t *testing.T) {
 			"argocd": api.PaasCapability{
 				CustomFields: map[string]string{
 					"git_revision": paasArgoGitRevision,
+					"git_url":      paasArgoGitUrl,
```
portly-halicore-76: Why do we need to add this? It is set via GitUrl already, line 46

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -158,3 +165,64 @@ func (v *PaasCustomValidator) validateSecrets(ctx context.Context, conf v1alpha1
...
+						Child("custom_fields"),
+					fname,
+					"custom field not configured",
+				))
+
+				continue
+			}
+
+			if c.CustomFields[fname] != "" && fconf.Validation != "" {
+				r, err := regexp.Compile(fconf.Validation)
```
portly-halicore-76: This should be caught by the webhook for the PaasConfig. Therefore we shouldn't do this everytime.

portly-halicore-76 (changes requested):

test/e2e/capabilityargocd_test.go
AxiomaticFixedChimpanzee: Because the test fixture set up in main_test.go specifically makes this a required custom field. Otherwise the test fails.

test/e2e/capabilityargocd_test.go
AxiomaticFixedChimpanzee: I was confused about the relationship between these two, but it again seems to be because it's configured as a required custom field in the test fixture. Without this the test fails.

internal/webhook/v1alpha1/paas_webhook.go
AxiomaticFixedChimpanzee: But we still need to compile the regex in order to match against it; the returned error is interpreted as an internal error which as you say should never be returned, but the alternative is to use `MustCompile()` which panics instead.

test/e2e/capabilityargocd_test.go
devotional-phoenix-97: e2e tests should not fail like this.
See https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/capabilities.go#L178 for logic in Paas controller.
Please reuse that implementation.

test/e2e/capabilityargocd_test.go
devotional-phoenix-97: e2e tests should not fail like this.
See https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/capabilities.go#L178 for logic in Paas controller.
Please reuse that implementation.

internal/webhook/v1alpha1/paas_webhook.go
devotional-phoenix-97: See https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/capabilities.go#L178 for logic in Paas controller.
Please reuse that implementation.

test/e2e/capabilityargocd_test.go
devotional-phoenix-97: See https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/capabilities.go#L178 for logic in Paas controller.
This relies on https://github.com/belastingdienst/opr-paas/blob/main/api/v1alpha1/paas_types.go#L338 which merges these direct fields with custom fields.
If you reuse that implementation, this custom field is set (from direct field) and therefore required check does not fail.

portly-halicore-76 (approved):
Very nice!

PR #364, opened 2025-02-12
feat: Paas webhook, group validation

portly-halicore-76:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

No validation for groups in a webhook

Fixes: # (issue)

## What is the new behavior?

Paas webhook warns when a group contains both users and a query. 

- [ ] Yes
- [x] No

### DIFF
--- a/internal/webhook/v1alpha1/paas_webhook.go
+++ b/internal/webhook/v1alpha1/paas_webhook.go
@@ -89,6 +89,7 @@ func (v *PaasCustomValidator) ValidateDelete(ctx context.Context, obj runtime.Ob
 
 func (v *PaasCustomValidator) validate(ctx context.Context, paas *v1alpha1.Paas) (admission.Warnings, error) {
 	var allErrs field.ErrorList
+	var warnings []string
 	conf := config.GetConfig()
 	// Check for uninitialized config
 	if conf.DecryptKeysSecret.Name == "" {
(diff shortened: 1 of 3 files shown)


internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -103,11 +104,13 @@ func (v *PaasCustomValidator) validate(ctx context.Context, paas *v1alpha1.Paas)
 		allErrs = append(allErrs, errs...)
 	}
 
-	if len(allErrs) == 0 {
+	warnings = append(warnings, v.validateGroups(paas.Spec.Groups)...)
+
+	if len(allErrs) == 0 && len(warnings) == 0 {
```
AxiomaticFixedChimpanzee, line 109: If there are no errors but there are warnings, we don't want to skip this return statement because otherwise we'll be returning an error when there isn't one.

AxiomaticFixedChimpanzee (changes requested):

internal/webhook/v1alpha1/paas_webhook_test.go
```diff
@@ -114,31 +114,77 @@ var _ = Describe("Paas Webhook", func() {
...
 				To(MatchError(ContainSubstring("capability not configured")))
 		})
 
-		// It("Should validate updates correctly", func() {
-		//     By("simulating a valid update scenario")
-		//     oldObj.SomeRequiredField = "updated_value"
-		//     obj.SomeRequiredField = "updated_value"
-		//     Expect(validator.ValidateUpdate(ctx, oldObj, obj)).To(BeNil())
-		// })
+		It("When updating a Paas should warn when a group contains both users and a query", func() {
```
devotional-phoenix-97: Should be rephrased. Maybe something like:
should generate a warning when updating a Paas with a group that contains both users and a queries,

internal/webhook/v1alpha1/paas_webhook_test.go
```diff
@@ -114,31 +114,77 @@ var _ = Describe("Paas Webhook", func() {
...
+						},
+					},
+				},
+			}
+
+			warnings, _ := validator.ValidateUpdate(ctx, nil, obj)
+			Expect(warnings).To(ContainElement("spec.groups[foo] contains both users and query, the users will be ignored"))
+		})
+
+		It("When updating a Paas should not warn when a group contains just users", func() {
```
devotional-phoenix-97: should not warn when updating a Paas with a group that contains just users

devotional-phoenix-97 (changes requested):

internal/webhook/v1alpha1/paas_webhook.go
portly-halicore-76: Like this?

AxiomaticFixedChimpanzee (dismissed):

devotional-phoenix-97 (approved):

PR #382, opened 2025-02-19
feat(paas_webhook): warn if quota requests higher than limits

### LINKED ISSUE #235: feat: Implement AdmissionWebhook for Paas

AxiomaticFixedChimpanzee:
Adds warnings for quota and extra permission discrepancies to the Paas admission webhook. Closes #235.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

- [ ] Yes
- [x] No

### DIFF
--- a/internal/webhook/v1alpha1/paas_webhook.go
+++ b/internal/webhook/v1alpha1/paas_webhook.go
@@ -14,6 +14,7 @@ import (
 	"github.com/belastingdienst/opr-paas/internal/config"
 	"github.com/belastingdienst/opr-paas/internal/crypt"
 	"github.com/belastingdienst/opr-paas/internal/logging"
+	"github.com/belastingdienst/opr-paas/internal/quota"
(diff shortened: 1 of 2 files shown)


portly-halicore-76 (approved):
Amazing

PR #407, opened 2025-02-26
fix(PaasConfigWebhook): do not check Min > Def error situation if Clusterwide

hikarukin:
When creating/updating a new PaasConfig having a capability defined with Clusterwide
set to true, it should not check if MinQuota > DefQuota

Signed-off-by: hikarukin <156204784+hikarukin@users.noreply.github.com>

## Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes

When Clusterwide is set on a capability, the validation webhook considers a given
PaasConfig invalid if the MinQuota is larger than the DefQuota despite the fact
that in a clusterwide setup, this can be a valid scenario.

## What is the new behavior?

The validation webhook only checks if MinQuota > DefQuota if Clusterwide is false.

- [ ] Yes
- [X] No


devotional-phoenix-97 (dismissed):

devotional-phoenix-97:
We need to rebase this against #409 or the other way around.

TazmanianDevil666 (approved):

PR #409, opened 2025-02-26
feat: configurable groupname verification

### LINKED ISSUE #408: Groupname verification with RFC-1035 is too tight.
# Bug Report

**Paas Operator version:**

v1.5.0

**Current behaviour:**

When createign Paas, the groupnames variable is too thight.
In our case we have underscores in naming, but might also have uppercase letters, etc.

**Expected behaviour:**

Would be awesome if we could influence the groupnames verification.

**Steps to reproduce:**

create a paas like:
```
apiVersion: cpet.belastingdienst.nl/v1alpha1
kind: Paas
metadata:
  name: mypaas
spec:
  groups:
    my_group:
      users:
        - john
        - jane
  quota:
    limits.cpu: '1'
  requestor: me
```

And you receive an error like:
> Error "Invalid value: "my_group": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name', or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')" for field "spec.groups".

**Related code:**

[code](https://github.com/belastingdienst/opr-paas/blob/v1.5.0/internal/webhook/v1alpha1/paas_webhook.go#L155-L161)

```
		for _, errString := range validation.IsDNS1035Label(name) {
			errs = append(errs, field.Invalid(
				field.NewPath("spec").Child("groups"),
				name,
				errString,
			))
		}
```

**Other information:**

I will issue a PR

devotional-phoenix-97:
## Pull Request type

- [x] Bugfix
- [x] Feature
- [ ] Build-related changes

Fixes: #408

## What is the new behavior?

- you can set a regexp in PaasConfig which will be used to verify the groupNames
- regexp will be verified by webhook on creation / updates of PaasConfig
- Paas will be verified by webhook on creation / updates of Paas

- [ ] Yes
- [x] No

## Other information

We should expand this feature (e.a. regexp on requestor could be one other validation).


devotional-phoenix-97:
We need to rebase this against #407 or the other way around.

internal/webhook/v1alpha1/paas_webhook_test.go
```diff
@@ -263,30 +290,17 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
-					strings.Repeat(validChars, 3):              tooLongMsg,
+				invalidNames = []string{
+					"_" + validChars,
+					"A" + validChars,
+					strings.Repeat(validChars, 3)[0:64],
 				}
 			)
-			for invalidName, msg := range invalidNames {
+			for _, invalidName := range invalidNames {
+				fmt.Fprintf(GinkgoWriter, "DEBUG - InvalidName: %s", invalidName)
```
hikarukin, line 303: Did you leave this in on purpose?

hikarukin (approved):

internal/webhook/v1alpha1/paas_webhook_test.go
devotional-phoenix-97: Yeah, I needed this info to understand which value was failing.
This is only displayed with go test -v.
We should probably add debugging comments like this more often in our ginkgo code...

PR #477, opened 2025-03-19
feat: templating module

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

We currently don;t do anything with templating

## What is the new behavior?

- With this feature we added a new templating module
- Templating module is not yet used (future PR's)

- [ ] Yes
- [x] No

### DIFF
--- a/internal/templating/main.go
+++ b/internal/templating/main.go
@@ -1,0 +1,54 @@
+package templating
+
+import (
+	"bytes"
+	"text/template"
+
+	api "github.com/belastingdienst/opr-paas/api/v1alpha1"
+)
+
+type Templater struct {
+	Paas   api.Paas
+	Config api.PaasConfig
+}
+
+func NewTemplater(paas api.Paas, config api.PaasConfig) Templater {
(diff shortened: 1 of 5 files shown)


internal/templating/main.go
```diff
@@ -0,0 +1,54 @@
...
+		Config: config,
+	}
+}
+
+func (t Templater) Verify(name string, templatedText string) error {
+	_, err := template.New(name).Parse(templatedText)
+	return err
+}
+
+func (t Templater) TemplateToString(name string, templatedText string) (string, error) {
```
hikarukin, line 27: I assume thise is meant as a convenience function for external templates rather than converting `t` into a string? Because if its intended as the latter, ToString would be more appropriate.

hikarukin (dismissed):
I'd like to have function level comments for every function.

Not sure if our linters cover that right now but they should.

I want to enforce function level comments for the entire repo.

hikarukin (approved):

PR #494, opened 2025-03-25
chore: Fixes #492 - Store entire PaasConfig for go-template feature

### LINKED ISSUE #492: feat: Store entire PaasConfig for go-template feature
# Feature Request

When we added go-template functionality, we decided to add PaasConfig too.
But the current PaasConfig watcher implementation only stores the PaasConfig Spec, and here we should use the entire PaasConfig block.
We now create a dummy PaasConfig, but that can bring unexpected behavior (as in .Config.Name returns a different value then the actual name).

We should change the Watcher code, so that the entire PaasConfig is stored instead of just the Spec block.
We could then have a new function to return the entire PaasConfig, and have the original function return the Spec.
We could the use the entire PaasConfig in the templater code, and have all else use the Spec block.

**Describe Alternatives**

- Properly describe (but is unexpected behavior)
- have the templater fetch it's own version of PaasConfig (duplicating code, functionality, extra api calls, etc.)

**Related Code**

https://github.com/belastingdienst/opr-paas/blob/86f2315eef4481b9e1d86b0c611a502b7ca63c06/internal/config/paas_config_store.go#L26

**Additional Context**

- [x] Yes
- [ ] No

devotional-phoenix-97:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

When we added go-template functionality, we decided to add PaasConfig too.
But the current PaasConfig watcher implementation only stores the PaasConfig Spec, and here we should use the entire PaasConfig block.
We now create a dummy PaasConfig, but that can bring unexpected behavior (as in .Config.Name returns a different value then the actual name).

## What is the new behavior?

Watcher code now stores the PaasConfig entirely.
Previous method is renamend for GetConfig to GetConfigSpec, amd all code uses that implementation.
A new function GetConfig returns the entire PaasConfig.
Templater code now has the complete PaasConfig including annotations, lables, and correct name.

- [ ] Yes
- [x] No

### DIFF
--- a/internal/config/paas_config_store.go
+++ b/internal/config/paas_config_store.go
@@ -16,22 +16,29 @@ import (
 
 // PaasConfigStore is a thread-safe store for the current PaasConfig
 type PaasConfigStore struct {
-	currentConfig v1alpha1.PaasConfigSpec
+	currentConfig v1alpha1.PaasConfig
 	mutex         sync.RWMutex
 }
 
 var cnf = &PaasConfigStore{}
 
-// GetConfig retrieves the current configuration
-func GetConfig() v1alpha1.PaasConfigSpec {
+// GetConfigSpec retrieVes the current configuration
+func GetConfig() v1alpha1.PaasConfig {
 	cnf.mutex.RLock()
 	defer cnf.mutex.RUnlock()
 	return cnf.currentConfig
+}
+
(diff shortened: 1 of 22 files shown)


portly-halicore-76 (approved):
GO

PR #515, opened 2025-04-03
chore(style): enable staticcheck linter

AxiomaticFixedChimpanzee:
Enable staticcheck linter and fix violations.

Disables the default no-deprecations error, which causes too many failures at the moment and requires refactoring.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -11,7 +11,7 @@ linters:
     - ineffassign
     - misspell
     - revive
-    # - staticcheck
+    - staticcheck
     - unused
   settings:
     revive:
@@ -66,6 +66,16 @@ linters:
         - name: superfluous-else
         - name: unchecked-type-assertion
         - name: unexported-naming
+    staticcheck:
+      checks:
+        - all
+        - '-ST1000'
+        - '-ST1003'
+        - '-ST1016'
+        - '-SA1019'
+        - '-ST1020'
+        - '-ST1021'
+        - '-ST1022'
   exclusions:
     generated: lax
     rules:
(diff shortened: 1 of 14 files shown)


internal/templating/main_test.go
```diff
@@ -163,25 +162,25 @@ func TestCapCustomFieldsToMap(t *testing.T) {
 {{end}}`
 	)
 	var (
-		paas = v1alpha1.Paas{
+		paas = api.Paas{
```
hikarukin: We need to discuss this for a moment to set a standard. Reason: I used to use "api" but got overruled to use "v1alpha1" instead. I'd prefer to use "api" so I'm in favor of this, but I prefer consistency even more.

portly-halicore-76: We're working with multiple versions in the near future, how will that impact?

.golangci.yaml
```diff
@@ -76,6 +86,7 @@ linters:
       - third_party$
       - builtin$
       - examples$
+      - cmd/crypttool/
```
portly-halicore-76: This is to be removed in the near near future right @hikarukin

internal/templating/main_test.go
AxiomaticFixedChimpanzee: We could use the default package import name (i.e. the version specifier) in files that reference multiple versions simultaneously, and keep `api` for convenience in files that don't (the vast majority I would think)? In any case I don't have a strong opinion on it.

portly-halicore-76: I'ld like to specify the api version in the code as we are working an multiversion api's. So `v1alpha1.Paas{}` and `v1alpha2.Paas{}` are valid.

hikarukin (approved):
LGTM

PR #537, opened 2025-04-14
feat: removing paasns reconciler

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

- PaasNs controller is taking care of most of the PaasNS reconciliation efforts
- There is overlapping methods between the PaasNS reconciler and Paas Reconciler
- Implementation requires the Paas field to be set (for it to be efficient at least)

## What is the new behavior?

- Paas controller is taking care of all of the PaasNS reconciliation efforts
- There is no code for PaasNS reconciler
- Implementation works without the Paas field, and is still efficient

- [ ] Yes
- [x] No

## Other information

This is prep work for v1alpha2 which has no paas field anymore


internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -109,6 +110,9 @@ type paasSpecValidator func(
 func (v *PaasCustomValidator) validate(ctx context.Context, paas *v1alpha1.Paas) (admission.Warnings, error) {
 	var allErrs field.ErrorList
 	var warnings []string
+	if paas.DeletionTimestamp != nil {
```
portly-halicore-76, line 113: This is caught by not webhook validate onDelete?

devotional-phoenix-97: During delete, the paas controller frequently updates the paas, which goes through the webhook too.
I ran into issues which was resolved by adding this.
After which I found out it improves delete performance and e2e performance bigtime too.
I checked and we can do without this block if we want to.
But we can also be more relaxed with webhook checks when the object is being deleted, which saves performance.

portly-halicore-76: true!

PR #560, opened 2025-05-08
ci: Improving of coverage on paas_controller

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

- We currently have bare minimum unittest coverage on internal/controller/paas_controller.go
- We have moved controller code from paasns controller to paas_controller and have no coverage on that either. We might have a degration in v1alpha2 which we are unaware of.
- We want to upgrade controller from managing v1alpha1 to managing v1alpha2, and would like to have improved coverage on internal/controller
- We want to use ginkgo / gomega for unittesting of controller logic

## What is the new behavior?

- We have on par unittest coverage on internal/controller/paas_controller.go
- We have insights in degradations and resolved them
- We are one step closer to upgrading controller from managing v1alpha1 to managing v1alpha2 with minimal toil
- We use ginkgo / gomega for unittesting of paas controller logic

- [ ] Yes
- [x] No

### DIFF
--- a/internal/controller/paas_controller.go
+++ b/internal/controller/paas_controller.go
@@ -93,7 +93,19 @@ func (r *PaasReconciler) getPaasFromRequest(
 
 	if paas.GetDeletionTimestamp() != nil {
 		logger.Info().Msg("paas marked for deletion")
-		return nil, r.updateFinalizer(ctx, paas)
+		if !controllerutil.ContainsFinalizer(paas, paasFinalizer) {
+			return nil, nil
+		}
+		for _, finalizationFunc := range []func(context.Context, *v1alpha1.Paas) error{
+			r.setFinalizing,
+			r.finalizePaas,
+			r.removeFinalizer,
+		} {
+			if err = finalizationFunc(ctx, paas); err != nil {
+				return nil, err
+			}
+		}
+		return nil, nil
 	}
 
 	// Started reconciling, reset status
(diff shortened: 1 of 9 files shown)


internal/controller/paas_controller.go
```diff
@@ -365,7 +339,23 @@ func (r *PaasReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&v1alpha1.PaasNS{},
 			handler.EnqueueRequestsFromMapFunc(
 				func(_ context.Context, obj client.Object) []reconcile.Request {
-					return paasFromPaasNs(mgr, obj)
+					var ns corev1.Namespace
+					logger := mgr.GetLogger()
+					if err := mgr.GetClient().Get(
+						context.Background(),
+						types.NamespacedName{Name: obj.GetNamespace()},
```
portly-halicore-76: adding ns to name?

internal/controller/paas_controller.go
```diff
@@ -365,7 +339,23 @@ func (r *PaasReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&v1alpha1.PaasNS{},
 			handler.EnqueueRequestsFromMapFunc(
 				func(_ context.Context, obj client.Object) []reconcile.Request {
-					return paasFromPaasNs(mgr, obj)
+					var ns corev1.Namespace
```
portly-halicore-76, line 343: Please split this handler into another PR. Why you ask? Because this will be one mega pr later to merge from alpha2 to main. When splitting this into separate pr's, we can relate them to to be created issues. In that way we can scope the pr's and issues and trace them later. This is related to changing the paasns to paas controller, which should be documented in a separate issue.

internal/controller/paas_controller.go
```diff
@@ -385,11 +375,12 @@ func (r *PaasReconciler) finalizePaas(ctx context.Context, paas *v1alpha1.Paas)
 	logger.Debug().Msg("inside Paas finalizer")
 
 	paasReconcilers := []func(context.Context, *v1alpha1.Paas) error{
```
portly-halicore-76, line 378: finalizing is only needed where cascading delete does not apply. It should be that (when setting the owner refs ok, cascading delete on all owned resources takes place) only if some logic needs to be executed before paas deletion, for example recalculating a clusterwide quota, a finalizing method is justified. Therefore, IMHO, finalizeClusterQuotas and finalizeNamespaces can be removed entirely from the project, reducing complexity. I haven't checked finalizePaasClusterRoleBindings but that might be able to go as well.

internal/controller/rolebindings_controller.go
```diff
@@ -223,8 +223,8 @@ func (r *PaasReconciler) reconcileNamespaceRolebindings(
 		// Convert the groupKey to a groupName to map the rolebinding subjects to a group
 		groupName := paas.GroupKey2GroupName(groupKey)
 		for _, mappedRole := range config.GetConfig().Spec.RoleMappings.Roles(groupRoles) {
-			if role, exists := roles[mappedRole]; exists {
-				roles[mappedRole] = append(role, groupName)
+			if groups, exists := roles[mappedRole]; exists {
```
portly-halicore-76, line 226: this was role and should be role as groupName is the latter part of the map.

debug.test927367961
portly-halicore-76, line 1: Don't commit this to the project pls.

portly-halicore-76 (dismissed):
Nice improvement to the project but left some remarks to cover first.

internal/controller/paas_controller.go
devotional-phoenix-97: This PR only has unittest stuff, but that are 2 things:
1: Adding unittests
2: changing code so that it is easier to add unittests.
This specific change is definitely the latter.

devotional-phoenix-97: The function already existed, but as a separate function which did 2 things:
- Manager stuff, which I moved to be part of the main SetupWithManager function
- Finding paas that a paasns belongs to, which I left where it was
By moving the manager specific stuff to the SetupWithManager code the 'finding paas' stuff was easier to unittest which was the main goal.

internal/controller/paas_controller.go
devotional-phoenix-97: Yes.
Basically, I read the namespace that the paasns obj belongs to grom k8s find the Paas OwnerReference from the Namespace and schedule that paas for reconciliation.

internal/controller/paas_controller.go
devotional-phoenix-97: We now have proper unittests, let's check if that is indeed how it works and remove it if we can ;)

internal/controller/rolebindings_controller.go
devotional-phoenix-97: Roles is a dict which holds all roles, and for every role, the groups that should have this role.
Therefore roles[mappedRole] returns 'the groups belonging to this role'.

debug.test927367961
devotional-phoenix-97: Removed...

internal/controller/paas_controller.go
portly-halicore-76: I've [deleted similar in the past](https://github.com/belastingdienst/opr-paas/pull/228), as this is default k8s behaviour. Once last ownerref on a resource is deleted through cascading delete, the resource will be deleted as well. This will be harder to test in EnvTest due to: https://book.kubebuilder.io/reference/envtest.html#testing-considerations

internal/controller/rolebindings_controller.go
portly-halicore-76: ah check

internal/controller/paas_controller.go
portly-halicore-76: AAAH, yeah of course.

internal/controller/paas_controller.go
devotional-phoenix-97: Let's work on this in a new PR

hikarukin (approved):

PR #588, opened 2025-05-20
feat: move to v1alpha2.PaasConfig using conversion for v1alpha1.PaasConfig

hikarukin:
Move to use v1alpha2.PaasConfig

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Code uses v1alpha1.PaasConfig

## What is the new behavior?

Code uses v1alpha2.PaasConfig

- [ ] Yes
- [ ] No

## Other information

Code uses ConvertTo() and ConvertFrom() to support both v1alpha1 as well as v1alpha2.


api/v1alpha1/paasconfig_conversion.go
```diff
@@ -0,0 +1,10 @@
+/*
+Copyright 2025, Tax Administration of The Netherlands.
+Licensed under the EUPL 1.2.
+See LICENSE.md for details.
+*/
+
+package v1alpha1
+
+// Hub marks this type as a conversion hub.
+func (*PaasConfig) Hub() {}
```
portly-halicore-76, line 10: Haven't seen this before. Otters used an annotation for this. Where did you find this func's behaviour? See: https://github.com/belastingdienst/opr-paas/blob/v1alpha2/api/v1alpha1/paas_types.go#L436

api/v1alpha2/paasconfig_conversion.go
```diff
@@ -0,0 +1,262 @@
...
+	spec.Capabilities = make(map[string]v1alpha1.ConfigCapability, len(pc.Spec.Capabilities))
+	for key, val := range pc.Spec.Capabilities {
+		spec.Capabilities[key] = val.ConvertTo()
+	}
+
+	spec.GroupSyncList = pc.Spec.GroupSyncList.ConvertTo()
+	spec.GroupSyncListKey = pc.Spec.GroupSyncListKey
+	spec.LDAP = v1alpha1.ConfigLdap(pc.Spec.LDAP)
+	spec.ArgoPermissions = v1alpha1.ConfigArgoPermissions{}
+	spec.ArgoEnabled = false
```
portly-halicore-76, line 52: I believe this should be in the defaulter funcs. See https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation

api/v1alpha1/paasconfig_conversion.go
hikarukin: I simply followed https://github.com/belastingdienst/opr-paas/blob/da845ed86015a8a012008c22b61808fec6bc1af2/api/v1alpha1/paas_conversion.go#L10

api/v1alpha2/paasconfig_conversion.go
hikarukin: @portly-halicore-76 You're probably right. Are we using the defaulter stuff somewhere already, because a quick look at the code doesn't tell me much. If not, I'd prefer to do that in a separate PR.

api/v1alpha1/paasconfig_conversion.go
portly-halicore-76: Nvm.

api/v1alpha2/paasconfig_conversion.go
portly-halicore-76: Will be fixed in a separate PR related to #592

api/v1alpha2/paasconfig_conversion.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func (crm ConfigRoleMappings) convertFrom(src v1alpha1.ConfigRoleMappings) ConfigRoleMappings {
+	if src == nil {
+		return nil
+	}
+
+	dst := make(ConfigRoleMappings, len(src))
+	for k, v := range src {
+		dst[k] = append([]string(nil), v...)
```
portly-halicore-76, line 241: Why is this a []string(nil)

api/v1alpha2/paasconfig_conversion_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var paasconfigExV1Alpha1 = &v1alpha1.PaasConfig{
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "config-example",
+		Namespace: "platform",
```
portly-halicore-76: PaasConfig is clusterwide so this may go.

internal/config/paas_config_store.go
```diff
@@ -12,26 +12,46 @@ import (
...
+func GetConfig() v1alpha2.PaasConfig {
 	cnf.mutex.RLock()
 	defer cnf.mutex.RUnlock()
-	return cnf.currentConfig
+
+	return cnf.store
+}
+
+// GetConfigV1 retrieves the current configuration as a v1alpha1.PaasConfig
+func GetConfigV1() (v1alpha1.PaasConfig, error) {
```
portly-halicore-76, line 35: We've upgraded the operator to v1alpha2, so why would we need this?

internal/controller/capabilities.go
```diff
@@ -104,12 +105,22 @@ func (r *PaasReconciler) ensureAppSetCap(
...
+	capConfig := spec.Capabilities[capName]
+	templatedElements, err := applyCustomFieldTemplates(capConfig.CustomFields, templater)
 	if err != nil {
 		return err
 	}
+
 	capability := paas.Spec.Capabilities[capName]
-	capElements, err := capability.CapExtraFields(myConfig.Spec.Capabilities[capName].CustomFields)
+
+	// FIXME(hikarukin): temporarily convert v2 to v1
```
portly-halicore-76, line 117: Is this because we haven't migrated the paas reconciler to v1alpha2?

internal/templating/main_test.go
```diff
@@ -155,49 +153,49 @@ func TestInValidTemplateToMap(t *testing.T) {
...
-g, my-group-2, role:admin
-`,
-		}
-	)
-	tpl := templating.NewTemplater(paas, paasConfig)
-	tplResults, err := tpl.CapCustomFieldsToMap("my-cap")
-	assert.NoError(t, err)
-	assert.Equal(t, expected, tplResults)
-}
+// func TestCapCustomFieldsToMap(t *testing.T) {
```
portly-halicore-76: Please add a comment and explain why this test is commented out.

portly-halicore-76 (changes requested):
Great PR. Bit hard to review as of its size. Which is nothing to do about. 2 things I can't understand:

1. Am I overlooking or are the internal/webhook/v1alpha2 web hooks missing for paasconfig?
2. I cannot fully understand the changes to templating. Why is it needed to support both versions and make it generic.

internal/config/paas_config_store.go
hikarukin: For the v1alpha1 webhook mostly, since we need to keep using that for now.

internal/controller/capabilities.go
hikarukin: Yes, I mostly focussed on moving PaasConfig use to v1alpha2 and tried to avoid other parts since I felt that'd be out of scope of the PR and otherwise the PR would be bigger.

internal/templating/main_test.go
hikarukin: Related code was (re)moved. This is now removed as well.

api/v1alpha2/paasconfig_conversion_test.go
hikarukin: Agreed, removed.

api/v1alpha2/paasconfig_conversion.go
hikarukin: This is a Go idiom to create a true copy of a slice: `[]string(nil)` is a nil slice of the right type.

I use `append([]string(nil), v...)` to create a new slice with the same contents as v, but with a new underlying array. Essentially, I want to ensure that dst[k] and src[k] point to independent slices.

portly-halicore-76:
Please register the webhooks for v1alpha2 in cmd/main.go to fix issues.

hikarukin:
> Great PR. Bit hard to review as of its size. Which is nothing to do about. 2 things I can't understand:
> 
>     1. Am I overlooking or are the internal/webhook/v1alpha2 web hooks missing for paasconfig?
> 
>     2. I cannot fully understand the changes to templating. Why is it needed to support both versions and make it generic.

1. Fixed.
2. Needed because templating stuff is called from both v1 and v2 code.

portly-halicore-76 (approved):
Very very very nice!

PR #594, opened 2025-05-21
chore: Adding update tests to paas reconcile unittests

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

- Updates on a Paas contained no unittests

## What is the new behavior?

- Updates on a Paas have unitests
- Issues are fixed

- [ ] Yes
- [x] No

### DIFF
--- a/api/v1alpha1/groupversion_info.go
+++ b/api/v1alpha1/groupversion_info.go
@@ -18,6 +18,8 @@ var (
 	// GroupVersion is group version used to register these objects
 	GroupVersion = schema.GroupVersion{Group: "cpet.belastingdienst.nl", Version: "v1alpha1"}
 
+	paasAPIVersion = GroupVersion.Group + "/" + GroupVersion.Version
+
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
 	SchemeBuilder = &scheme.Builder{GroupVersion: GroupVersion}
 
--- a/api/v1alpha1/paas_types_test.go
+++ b/api/v1alpha1/paas_types_test.go
@@ -38,27 +38,29 @@ const (
 	ciCap        = "tekton"
 	ssoCap       = "keycloak"
 	extraCap     = "extra"
-	myKind       = "MyKind"
+	myKind       = "Paas"
 	otherKind    = "MyOtherKind"
-	myVersion    = "1.1.1"
 	otherVersion = "1.1.0"
 	myName       = "Some Name"
 	otherName    = "Some Other Name"
 )
 
-var testGroups = PaasGroups{
-	cntest1: PaasGroup{
-		Query: cntest2 + ",OU=org_unit,DC=example,DC=org",
-		Users: []string{"usr1", "usr2"},
-	},
-	cntest3: PaasGroup{
-		Query: cntest4,
-		Users: []string{"usr3", "usr2"},
-	},
-	tstGroup: PaasGroup{
-		Users: []string{"usr3", "usr2"},
-	},
-}
+var (
+	testGroups = PaasGroups{
+		cntest1: PaasGroup{
+			Query: cntest2 + ",OU=org_unit,DC=example,DC=org",
+			Users: []string{"usr1", "usr2"},
+		},
+		cntest3: PaasGroup{
+			Query: cntest4,
+			Users: []string{"usr3", "usr2"},
+		},
+		tstGroup: PaasGroup{
+			Users: []string{"usr3", "usr2"},
+		},
+	}
+	myVersion = paasAPIVersion
+)
 
 // Paas
 
--- a/internal/controller/capabilities.go
+++ b/internal/controller/capabilities.go
@@ -58,12 +58,11 @@ func (r *PaasReconciler) ensureAppSetCaps(
 	paas *v1alpha1.Paas,
 ) error {
 	paasConfigSpec := config.GetConfig().Spec
-	for capName := range paas.Spec.Capabilities {
+	for capName, capability := range paas.Spec.Capabilities {
 		if _, exists := paasConfigSpec.Capabilities[capName]; !exists {
 			return errors.New("capability not configured")
 		}
 		// Only do this when enabled
-		capability := paas.Spec.Capabilities[capName]
 		if enabled := capability.IsEnabled(); enabled {
 			if err := r.ensureAppSetCap(ctx, paas, capName); err != nil {
 				return err
@@ -207,14 +206,27 @@ func (r *PaasReconciler) finalizeAppSetCap(
 	return r.Patch(ctx, as, patch)
 }
 
-// finalizeAppSetCaps ensures all capabilities that exist are removed from the Capability Applicationsets
-func (r *PaasReconciler) finalizeAppSetCaps(
+// finalizeAllAppSetCaps removes this paas from all capability appsets
+func (r *PaasReconciler) finalizeAllAppSetCaps(
+	ctx context.Context,
+	paas *v1alpha1.Paas,
+) error {
+	paasWithoutCaps := paas.DeepCopy()
+	paasWithoutCaps.Spec.Capabilities = nil
+	return r.finalizeDisabledAppSetCaps(ctx, paasWithoutCaps)
+}
+
+// finalizeAppSetCaps removes this paas from all capability appsets that are not enabled in this paas
+func (r *PaasReconciler) finalizeDisabledAppSetCaps(
 	ctx context.Context,
 	paas *v1alpha1.Paas,
 ) error {
 	ctx, logger := logging.GetLogComponent(ctx, "Applicationsets")
 	for capName := range config.GetConfig().Spec.Capabilities {
 		logger.Info().Msgf("reconciling %s Applicationset", capName)
+		if capability, exists := paas.Spec.Capabilities[capName]; exists && capability.Enabled {
+			continue
+		}
 		err := r.finalizeAppSetCap(ctx, paas.Name, capName)
 		if err != nil {
 			return err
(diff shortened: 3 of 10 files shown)


api/v1alpha1/groupversion_info.go
```diff
@@ -18,6 +18,8 @@ var (
 	// GroupVersion is group version used to register these objects
 	GroupVersion = schema.GroupVersion{Group: "cpet.belastingdienst.nl", Version: "v1alpha1"}
 
+	paasAPIVersion = GroupVersion.Group + "/" + GroupVersion.Version
```
portly-halicore-76, line 21: I do feel like we need this in v1alpha2 as well?

api/v1alpha1/paas_types_test.go
```diff
@@ -38,27 +38,29 @@ const (
...
+		},
+		cntest3: PaasGroup{
+			Query: cntest4,
+			Users: []string{"usr3", "usr2"},
+		},
+		tstGroup: PaasGroup{
+			Users: []string{"usr3", "usr2"},
+		},
+	}
+	myVersion = paasAPIVersion
```
portly-halicore-76, line 62: Same here, needed in v1alpha2 as well?

internal/controller/capabilities.go
```diff
@@ -177,14 +176,27 @@ func (r *PaasReconciler) finalizeAppSetCap(
 	return r.Patch(ctx, as, patch)
 }
 
-// finalizeAppSetCaps ensures all capabilities that exist are removed from the Capability Applicationsets
-func (r *PaasReconciler) finalizeAppSetCaps(
+// finalizeAllAppSetCaps removes this paas from all capability appsets
```
portly-halicore-76, line 209: + by deliberately setting all capabilities to nil.

internal/controller/namespaces.go
```diff
@@ -120,3 +121,33 @@ func (r *PaasReconciler) reconcileNamespaces(
 	}
 	return nil
 }
+
+// finalizeObsoleteNamespaces returns all groups owned by the specified Paas
+func (r *PaasReconciler) finalizeObsoleteNamespaces(
```
portly-halicore-76, line 126: Rename please, deleteObsoleteNamespaces. Finalize is something else.

internal/controller/namespaces.go
```diff
@@ -120,3 +121,33 @@ func (r *PaasReconciler) reconcileNamespaces(
...
+		if _, exists := nsDefs[ns.Name]; exists {
+			continue
+		}
+		err = r.Delete(ctx, &ns)
+		if err != nil {
+			return err
+		}
+		i++
+	}
+	logger.Debug().Msgf("found %d existing groups owned by Paas %s", i, paas.Name)
```
portly-halicore-76: groups --> namespaces

internal/controller/namespaces.go
```diff
@@ -120,3 +121,33 @@ func (r *PaasReconciler) reconcileNamespaces(
...
+	logger := log.Ctx(ctx)
+	listOpts := []client.ListOption{
+		client.MatchingLabels(map[string]string{ManagedByLabelKey: paas.Name}),
+	}
+	err = r.List(ctx, &nss, listOpts...)
+	if err != nil {
+		return err
+	}
+	for _, ns := range nss.Items {
+		if _, exists := nsDefs[ns.Name]; exists {
```
portly-halicore-76: flip this, if !exists do the logic.

api/v1alpha1/groupversion_info.go
hikarukin: Especially in v1alpha2.

internal/controller/paas_controller.go
```diff
@@ -219,38 +219,48 @@ func (r *PaasReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
...
-		return ctrl.Result{}, err
+		return err
 	}
 	logger.Debug().Msgf("Need to create resourced for %d namespaces", len(nsDefs))
 	paasNsReconcilers := []func(context.Context, *v1alpha1.Paas, namespaceDefs) error{
 		r.reconcileNamespaces,
 		r.reconcilePaasRolebindings,
 		r.reconcilePaasSecrets,
 		r.reconcileClusterRoleBindings,
+		r.finalizeObsoleteNamespaces,
```
portly-halicore-76, line 254: Would like to see this as a call from the r.reconcileNamespaces func.

internal/controller/paas_controller.go
```diff
@@ -219,38 +219,48 @@ func (r *PaasReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
 	paasReconcilers := []func(context.Context, *v1alpha1.Paas) error{
 		r.reconcileQuotas,
 		r.reconcileClusterWideQuota,
+		r.reconcileNamespacedResources,
 		r.reconcileGroups,
 		r.ensureLdapGroups,
+		r.ensureAppSetCaps,
+		r.finalizeDisabledAppSetCaps,
```
portly-halicore-76, line 226: Would like to see this as a call from r.ensureAppSetCaps

portly-halicore-76 (changes requested):
Nice PR, please find my attached comments for a little improvement.

api/v1alpha1/groupversion_info.go
devotional-phoenix-97: Yes, it does, but not in this PR.
It should be added to v1alpha2 in the PR that migrates the controller code to use v1alpha2 resources over v1alpha1 resources...

api/v1alpha1/paas_types_test.go
devotional-phoenix-97: Again, yes, but not in this PR.
Let's merge this PR, so we can start working on migrating from controllers relying on v1alpha1 to relying on v1alpha2.

internal/controller/capabilities.go
devotional-phoenix-97: Is it important to state how it actually does it?
By the way, then it should say 'by copying the struct, removing all capabilities from the copy and finalizing the copy'...
I don't think the implementation (the how) should be documented in the comment, only the functionality (the what).

internal/controller/namespaces.go
devotional-phoenix-97: Looking at https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#external-resources I see
```
Sometimes external resources or resources that are not owned by a custom resource, those across namespaces for example, need to be cleaned up when the parent resource is deleted. In that case [Finalizers](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers) can be leveraged.
```
True, the parent is not deleted, it is modified.
And it is not the Paas finalizer, but the ObsoleteNamespace finalizer.

devotional-phoenix-97: When a sub resource (e.a. namespace in the ns block of a Paas) is deleted,
to me writing code to manage that and calling it a finalizers makes perfect sense.

devotional-phoenix-97: If you feel it must be named deleteObsoleteNamespaces, I will rename it, but my vote is for finalizeObsoleteNamespaces.

internal/controller/namespaces.go
devotional-phoenix-97: Fixed, thanks...

internal/controller/namespaces.go
devotional-phoenix-97: Sure.

internal/controller/paas_controller.go
devotional-phoenix-97: I thought about that too, but it feels a bit obfuscating to delete in some of the reconciliation functions.
And it isn't great for unit testing either. I'd rather keep it explicit like this...
Btw, we should change the logger output on line 235 to reflect that we are deleting here too.

internal/controller/paas_controller.go
devotional-phoenix-97: Again, keeping the two separated is cleaner (explicit and better for unit testing).
So let's not merge the two.

internal/controller/namespaces.go
portly-halicore-76: My point here is, as stated by the docs you ref, finalizing is doing manual things in the process of deleting a parent. For me, as stated in the other comment, deletingObsoleteNamespaces is part of the reconcilation process. We're trying to achieve the desired state as described by the current state of the Paas. That includes creating namespaces which aren't here yet, and deleting ones that have become obsolete. That is part of the reconcilation loop, not the finalization loop. Therefore, I believe, the func names should reflect that as well and preferably show the method chain by calling deleteObsoleteX within the reconcileX func.

portly-halicore-76: Ps. we have the same logic in place in [reconcileQuotas](https://github.com/belastingdienst/opr-paas/blob/v1alpha2/internal/controller/cluster_quotas.go#L162), groups and secrets.

internal/controller/paas_controller.go
portly-halicore-76: See my previous comment, have this setup in place in other reconcilers as well.

internal/controller/paas_controller.go
portly-halicore-76: Resolving this as will be part of the other conversations.

internal/controller/paas_controller.go
portly-halicore-76: As discussed, we postpone this for a next pr.

internal/controller/namespaces.go
portly-halicore-76: As discussed, we postpone this for a next pr.

portly-halicore-76 (approved):
Very nice.

Issue #595, opened 2025-05-23
feat: Merge RB entries


portly-halicore-76:
# Feature Request

**Describe the Feature Request**

It is possible that multiple equal entries end up in a rolebinding. This happens when there are multiple groups defined in a Paas with the same query value but a different groupKey. This is functionally unusually, but we must be able to deal with this.

Example paas groups entry:

```  
....  
  groups:
    - test_test:
         query: CN=tester
    - test_test2:
         query: CN=tester
```

Resulting in the following rolebinding entry:

```
.....
subjects:
  - kind: Group
    apiGroup: rbac.authorization.k8s.io
    name: tester
  - kind: Group
    apiGroup: rbac.authorization.k8s.io
    name: tester
```

**Describe Preferred Solution**

Merge the same entries in a rolebinding to one entry.

**Describe Alternatives**

X

**Related Code**

Assumable be here somewhere:

https://github.com/belastingdienst/opr-paas/blob/main/internal/controller/rolebindings_controller.go#L188-L199

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

portly-halicore-76:
Closed by https://github.com/belastingdienst/opr-paas/pull/603

PR #603, opened 2025-05-27
fix: Unique rolebinding entries

portly-halicore-76:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

When multiple groups with the same name are derived from the paas.groups spec, a rolebinding with duplicate entries will be created. 

Fixes: #595 

## What is the new behavior?

All entries in a rolebinding created by Paas, are unique combinations of groups and roles.

- [ ] Yes
- [x] No

### DIFF
--- a/internal/controller/rolebindings_controller.go
+++ b/internal/controller/rolebindings_controller.go
@@ -207,11 +207,11 @@ func (r *PaasReconciler) reconcileNamespaceRolebindings(
 ) error {
 	ctx, logger := logging.GetLogComponent(ctx, "rolebinding")
 	// Use a map of sets to avoid duplicates
-	roleGroups := map[string]map[string]bool{}
+	roleGroups := map[string]map[string]struct{}{}
 
 	for _, roleList := range config.GetConfig().Spec.RoleMappings {
 		for _, role := range roleList {
-			roleGroups[role] = map[string]bool{}
+			roleGroups[role] = map[string]struct{}{}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/controller/rolebindings_controller.go
```diff
@@ -206,14 +206,16 @@ func (r *PaasReconciler) reconcileNamespaceRolebindings(
 	nsName string,
 ) error {
 	ctx, logger := logging.GetLogComponent(ctx, "rolebinding")
-	roles := map[string][]string{}
+	// Use a map of sets to avoid duplicates
+	roleGroups := map[string]map[string]bool{}
```
AxiomaticFixedChimpanzee: If you use `map[string]struct{}` it acts like a set but doesn't take any extra memory to store a boolean :)

internal/controller/rolebindings_controller.go
```diff
@@ -223,13 +225,20 @@ func (r *PaasReconciler) reconcileNamespaceRolebindings(
...
+
+	roles := map[string][]string{}
+	for role, groupSet := range roleGroups {
+		for groupName := range groupSet {
+			roles[role] = append(roles[role], groupName)
+		}
+	}
+
 	logger.Info().Any("Rolebindings map", roles).Msg("creating paas RoleBindings for PAASNS object")
 	for roleName, groupNames := range roles {
```
AxiomaticFixedChimpanzee: This loop now seems redundant, since it's identical to the new outer loop we created above. We could just call `reconcileNamespaceRolebinding()` at line 240 instead, unless the log message with the complete map structure really needs to be logged prior to the reconciliation.

AxiomaticFixedChimpanzee (dismissed):

internal/controller/rolebindings_controller.go
portly-halicore-76: How was it than possible the "weird" duplicate behaviour was possible?

AxiomaticFixedChimpanzee: Oh, I meant the _inner_ set should be a map[string]struct{}. I.e.
```
roleGroups := map[string]map[string]struct{}
```
Then you have a map of sets, just as you have done, it's just that the sets are implemented as `map[string]struct{}` instead of `map[string]bool`.

AxiomaticFixedChimpanzee (approved):

PR #653, opened 2025-06-27
fix: Updates project to allow configuring metrics endpoint

### LINKED ISSUE #654: bug: Metrics can't be monitored
# Bug Report

**Paas Operator version:**

v2.1.0

**Current behaviour:**

EOF tls error in the logs of the operator. Investigating the root cause was that the metrics endpoint is now https, but not configured in the operator, thus scraping was no longer possible.

**Expected behaviour:**

Provide a working metric endpoint which can be configured and scraped. 

**Steps to reproduce:**

**Related code:**

```
insert short code snippets here
```

**Other information:**

This has to do with controller-runtime being updated and their way to deal with metrics endpoint security. See: 
https://book.kubebuilder.io/reference/metrics

portly-halicore-76:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Metrics endpoint can't be scraped. This has to do with a change in controller-runtime. We update our deps via Dependabot, but don't update boilerplate code provided by Kubebuilder.

Fixes: #654 

## What is the new behavior?

Updates the project according to latest controller-runtime setup in order to set have control over the metrics endpoint.

- [ ] Yes
- [x] No


nexus-core174 (dismissed):

manifests/rbac/metrics_reader_role.yaml
nexus-core174, line 1: This does not have a corresponding CRB in this deployment, not sure if that's on purpose

nexus-core174 (approved):

manifests/rbac/metrics_reader_role.yaml
portly-halicore-76: That is as designed. The role needs to be bind to the user scraping the metrics. Therefore, we don't know which user that will be. Implementors of Paas will need to create the rolebinding themselfs. This setup is an exact copy of the way Kubebuilder scaffolds an operator project.

PR #661, opened 2025-07-03
hotfix: Adding deprecation messages for unordered namespace lists and disabled capabilities

### LINKED ISSUE #658: bug: v1alpha1 <> v1alpha2 conversion not consistent
# Bug Report

## Paas Operator version

v2.x / v3.0.0

## Current behaviour

Issue is worst with ArgoCD. When ArgoCD creates a v1alpha1 Paas, it is written to etcd as a v1alpha2 paas,
and converted back to v1alpha1 when ArgoCD checks if all changes have been applied properly.
There are options that can be defined in v1alpha1, but not in v1alpha2. 
Functionally, they make no sense, but ArgoCD expects them when checking that all is applied succesfully and wines that they do not seem so.
few examples:

### example 1

```yaml
apiVersion: cpet.belastingdienst.nl/v1alpha1
kind: Paas
metadata:
  name: tst-tst
spec:
  capabilities:
    grafana:
      enabled: false
...
```
The above example would not have any effect on(grafana capability would not be created), but when converting to v1alpha2 and back to v1alpha1 would remove the grafana capability.

### example 2
```yaml
apiVersion: cpet.belastingdienst.nl/v1alpha1
kind: Paas
metadata:
  name: tst-tst
spec:
  namespaces:
  - ns2
  - ns1
...
```
The above example (reverse order of namespaces) would not have any effect on (both namespaces created, nothing special).
But when converting to v1alpha2 and back to v1alpha1 the order of the list would be random (maps are unordered, so storing v1alpha2.Paas would loose ordering info, and recreating v1alpha1 could have any order).
We could change order to be sorted when converting back to v1alpha1 (which we should), but this specific Paas would not be exactly as ArgoCD tried to create is (with ns2 before ns1 in list)...

## Expected behaviour

- ideally, we could store both versions and preserve all intermediate info of all versions, but that option is not available
- if we warn for capabilities with `enabled: false` we at least can check which ones should be fixed
- if we check and warn on unordered namespace block, and order properly when converting back it would already make ArgoCD happy

## Steps to reproduce

Install operator on cluster and create above examples with argocd, and argocd will keep mentioning changes on example 1 and flip between ok and mentioning changes on example 2

## Related code

Few pointers to code:
- [we could add a validation here](https://github.com/belastingdienst/opr-paas/blob/v3.0.0/internal/webhook/v1alpha1/paas_webhook.go#L211)
- [we already have the sort](https://github.com/belastingdienst/opr-paas/blob/v3.0.0/api/v1alpha1/paas_conversion.go#L76) and need to see why it still is seen as unsorted by ArgoCD
- [we could reverse these namespaces](https://github.com/belastingdienst/opr-paas/blob/v3.0.0/api/v1alpha1/paas_conversion_test.go#L59) and then [we could add a unittest to convert from and back and check that order is enforced](https://github.com/belastingdienst/opr-paas/blob/v3.0.0/api/v1alpha1/paas_conversion_test.go#L160)

devotional-phoenix-97:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Fixes: #658

## What is the new behavior?

We now have deprecation warnings for these edgecases (unordered namespace lists and disabled capabilities)

- [ ] Yes
- [x] No


internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -409,3 +417,29 @@ func (v *PaasCustomValidator) validateExtraPerm(conf v1alpha1.PaasConfig, paas *
 
 	return warnings
 }
+
+// validatePaasNamespaceListSorted returns an error when list is not sorted.
```
portly-halicore-76:
```suggestion
// validateListSorted returns a warning when the list is not sorted.
```

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -409,3 +417,29 @@ func (v *PaasCustomValidator) validateExtraPerm(conf v1alpha1.PaasConfig, paas *
...
+func validateListSorted(
+	list []string,
+) (warnings []string) {
```
portly-halicore-76, line 425: Could be a string right?

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -409,3 +417,29 @@ func (v *PaasCustomValidator) validateExtraPerm(conf v1alpha1.PaasConfig, paas *
...
+		return list[i] < list[j]
+	}) {
+		warnings = append(warnings, orderedListWarning)
+	}
+
+	return warnings
+}
+
+// validatePaasNamespaceListSorted returns an error when list is not sorted.
+func validateDIsabledCapabilities(
```
portly-halicore-76:
```suggestion
func validateDisabledCapabilities(
```

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -409,3 +417,29 @@ func (v *PaasCustomValidator) validateExtraPerm(conf v1alpha1.PaasConfig, paas *
...
+	if !sort.SliceIsSorted(list, func(i, j int) bool {
+		return list[i] < list[j]
+	}) {
+		warnings = append(warnings, orderedListWarning)
+	}
+
+	return warnings
+}
+
+// validatePaasNamespaceListSorted returns an error when list is not sorted.
```
portly-halicore-76:
```suggestion
// validateDisabledCapabilities .......
```

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -409,3 +417,29 @@ func (v *PaasCustomValidator) validateExtraPerm(conf v1alpha1.PaasConfig, paas *
...
+func validateDIsabledCapabilities(
+	capabilities v1alpha1.PaasCapabilities,
+) (warnings []string) {
```
portly-halicore-76, line 439: Same, just one string.

internal/webhook/v1alpha1/paas_webhook_test.go
```diff
@@ -155,10 +155,27 @@ var _ = Describe("Paas Webhook", Ordered, func() {
 				}
 			}
 		})
+		/*
```
portly-halicore-76: Why is this commented?

portly-halicore-76 (changes requested):
Few nitpicks!

internal/webhook/v1alpha1/paas_webhook.go
devotional-phoenix-97: Could be, but then we need code to verify if it is emptystring, (or nilpointer if you want *string) and append if it is not. This is easier and cleaner on the calling side...

internal/webhook/v1alpha1/paas_webhook_test.go
devotional-phoenix-97: Fixed

hikarukin (dismissed):

portly-halicore-76 (approved):
LGTM

PR #680, opened 2025-07-21
feat: disable userblock (webhook changes)

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

- we always allow a user block in groups
- we warn when both group and query are set

## What is the new behavior?

- we warn when both group and query are set
- we warn when feature flag is set to warn and group contains users
- we block when feature flag is set to block and group contains users

- [ ] Yes
- [x] No


portly-halicore-76:
I believe we should not only validate on feature flags but change operator behavior as well.

portly-halicore-76 (approved):
LGTM!

PR #681, opened 2025-07-21
chore: Removing revive:disable:exported

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

- We have exported api methods without comments

## What is the new behavior?

- some api methods have been unexported
- some api methods have been removed
- all remaining api methods have comments

- [x] Yes
- [ ] No

## Other information

By unexporting / removing methods that are no loinger used by this code base, we might break code in other codebases using this codebase


api/v1alpha1/paasns_types.go
```diff
@@ -131,12 +91,7 @@ type PaasNsStatus struct {
 
 // revive:enable:line-length-limit
 
-// Deprecated: use paasns.status.conditions instead
-func (ps *PaasNsStatus) Truncate() {
+// Truncate is Deprecated: use paasns.status.conditions instead
```
portly-halicore-76, line 94: I believe all these changes should be two liners. One as it were, `Depracted: ...` The other describing the original func behavior. Without Deprecated being first in the sentence, Ideas will not warn when using depcrated funcs.

portly-halicore-76 (changes requested):

api/v1alpha1/paasns_types.go
devotional-phoenix-97: revive:disable:exported requires a format such as:
[Object name] ....
- Must start with name of the struct, function, etc.
- Must not end with '.'
- Can be multiline, but then first line must start with ...

So starting with 'Deprecated' raises linting errors where revive:disable:exported is enabled.

PR #696, opened 2025-08-11
feat: Implements plugin generator for ArgoCD

portly-halicore-76:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Integration for capabilities with ArgoCD is done via ListGenerator manipulation

## What is the new behavior?

In addition to the ListGenerator manipulation, there is a pluginGenerator server available. This can be enabled to provide a pull-based mechanism to integrate with ArgoCD. This will replace the ListGenerator way in the next major.

- [ ] Yes
- [x] No


cmd/manager/main.go
```diff
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	argocdplugingenerator "github.com/belastingdienst/opr-paas/v3/internal/argocd-plugin-generator"
```
nexus-core174, line 18: The linter complained about underscores in the name before right? Can you do it with dashes instead? I feel like that would be the best. `argocd-plugin-generator`

nexus-core174 (dismissed):

cmd/manager/main.go
portly-halicore-76: dashes are not allowed

nexus-core174 (approved):

PR #721, opened 2025-08-21
feat: Quota validation

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

## What is the new behavior?

- You can now set validations in PaasConfig to limit the list of valid Quota names
- If Quota Name validations is set, PaasConfig capability quota's must meet the validations
- If Quota Name validations is set, Paas capability quota's must meet the validations
- If Quota Name validations is set, Paas quota's must meet the validations

- [ ] Yes
- [x] No

## Other information

Docs will be a new PR


internal/webhook/v1alpha1/paas_webhook_test.go
```diff
@@ -489,6 +489,80 @@ var _ = Describe("Paas Webhook", Ordered, func() {
 			//revive:disable-next-line
 			Expect(warnings[0]).To(Equal("spec.capabilities[bar].extra_permissions capability does not have extra permissions configured"))
 		})
+		Context("quota name validation", func() {
+			var (
+				validResourceKeys = []string{
+					// "limits.cpu",
```
portly-halicore-76, line 495: Pls remove this comment

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -215,6 +216,41 @@ func validatePaasName(
...
 
+// validatePaasQuotaNames returns errors if there are quota's with keys that do not meet validations
+func validatePaasQuotaNames(
+	_ context.Context,
+	_ client.Client,
+	conf v1alpha1.PaasConfig,
+	paas *v1alpha1.Paas,
+) ([]*field.Error, error) {
+	var errs []*field.Error
+	nameValidationRE := conf.Spec.Validations.GetValidationRE("paas", "quotaNames")
```
portly-halicore-76: Isn't: "allowedQuotas" or as the upstream calls it: "corev1.ResourceName" something a more appropriate flag to set?

internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -215,6 +216,41 @@ func validatePaasName(
...
+		quotas[cf.Key(name).Child("quota")] = c.Quota
+	}
+
+	for f, q := range quotas {
+		for quotaKey := range q {
+			if !nameValidationRE.Match([]byte(quotaKey)) {
+				errs = append(errs, field.Invalid(
+					f,
+					quotaKey,
+					fmt.Sprintf("quota key does not match configured validation regex `%s`", nameValidationRE.String()),
```
portly-halicore-76: I believe something like: "requested resourceName is not allowed". An end-user requesting a Paas doesn't want to see the regex returned but needs to know something is not allowed IMHO.

internal/webhook/v1alpha1/paas_webhook_test.go
```diff
@@ -489,6 +489,80 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+				}
+				validation       = fmt.Sprintf("(%s)", strings.Join(validResourceKeys, "|"))
+				validationConfig = v1alpha1.PaasConfigValidations{
+					"paas": v1alpha1.PaasConfigTypeValidations{"quotaNames": validation},
+				}
+				invalidQuotas = map[corev1.ResourceName]resource.Quantity{
+					"limits.cpu": resource.MustParse("1.1"),
+				}
+			)
+			It("should allow cap quota names that meet re", func() {
```
portly-halicore-76, line 517: can you add a testcase, in which a cap is configured without a Quota.

internal/webhook/v1alpha2/paas_webhook.go
```diff
@@ -210,6 +211,41 @@ func validatePaasName(
...
+		for quotaKey := range q {
+			if !nameValidationRE.Match([]byte(quotaKey)) {
+				errs = append(errs, field.Invalid(
+					f,
+					quotaKey,
+					fmt.Sprintf("quota key does not match configured validation regex `%s`", nameValidationRE.String()),
+				))
+			}
+		}
+	}
```
portly-halicore-76: Can you extract this to a generic file to reuse between v1alpha1 and v1alpha2 to prevent duplicate code / maintenance?

internal/webhook/v1alpha2/paasconfig_webhook.go
```diff
@@ -273,6 +285,39 @@ func validateConfigQuotaSettings(qs v1alpha2.ConfigQuotaSettings, rootPath *fiel
...
 }
 
+func validateQuotaNames(
+	qs v1alpha2.ConfigQuotaSettings,
+	validations v1alpha2.PaasConfigValidations,
+	childPath *field.Path,
+) field.ErrorList {
+	var allErrs field.ErrorList
+	childPath = childPath.Child("quotas")
+	// We use same value for paas.spec.namespaces and paasns.metadata.name validation.
```
portly-halicore-76: Pls remove unrelated comment.

internal/webhook/v1alpha2/paasconfig_webhook.go
```diff
@@ -273,6 +285,39 @@ func validateConfigQuotaSettings(qs v1alpha2.ConfigQuotaSettings, rootPath *fiel
...
+	// Unless both are set.
+	nameValidationRE := validations.GetValidationRE("paas", "quotaNames")
```
portly-halicore-76: Same as before, pls find a more suitable regexName

internal/webhook/v1alpha2/paasconfig_webhook.go
```diff
@@ -273,6 +285,39 @@ func validateConfigQuotaSettings(qs v1alpha2.ConfigQuotaSettings, rootPath *fiel
...
+		"min":      qs.MinQuotas,
+		"max":      qs.MaxQuotas,
+	} {
+		for quotaKey := range quotas {
+			if !nameValidationRE.Match([]byte(quotaKey)) {
+				allErrs = append(allErrs, field.Invalid(
+					childPath.Child(kind),
+					quotaKey,
+
+					fmt.Sprintf("quota key does not match configured validation regex `%s`", nameValidationRE.String()),
```
portly-halicore-76: Some as before, pls add a more descriptive err message.

internal/webhook/v1alpha2/paasconfig_webhook.go
```diff
@@ -273,6 +285,39 @@ func validateConfigQuotaSettings(qs v1alpha2.ConfigQuotaSettings, rootPath *fiel
...
+				))
+			}
+		}
+	}
+
+	return allErrs
+}
+
```
portly-halicore-76: 99% of this is an exact copy for v1alpha1, can we extract this and call that from both v1 and v2?

internal/webhook/v1alpha2/paasconfig_webhook_test.go
```diff
@@ -280,6 +281,59 @@ var _ = Describe("Creating a PaasConfig", Ordered, func() {
...
+				}
+			)
+			It("should allow names that meet re", func() {
+				obj.Spec.Validations["paas"]["quotaNames"] = validation
+				obj.Spec.Capabilities = v1alpha2.ConfigCapabilities{
+					"quota": v1alpha2.ConfigCapability{
+						QuotaSettings: v1alpha2.ConfigQuotaSettings{
+							DefQuota:  validQuotas,
+							MinQuotas: validQuotas,
+							MaxQuotas: validQuotas,
```
portly-halicore-76, line 313: What is one of these is not set? Will that break the code?

portly-halicore-76 (dismissed):
Thanks for this rather helpfull PR! I've left a few nitpicks to discuss.

internal/webhook/v1alpha1/paas_webhook_test.go
devotional-phoenix-97: I have that in there on purpose, so that everyone sees that this is the one we don;t have in the list of approved resource names...

internal/webhook/v1alpha1/paas_webhook.go
devotional-phoenix-97: allowedQuotas seems good.
I will update...

internal/webhook/v1alpha1/paas_webhook.go
devotional-phoenix-97: 1. I don't want to call is resourceName, as (to me at least) it sounds like the name of the Paas itself is the issue...
2. Why not having RE in warning? It helps end user to know what is allowed, and what is not, right?

I will rephrase to something like:
quota %s is not allowed in PaasConfig, but will leave RE in there for now...

internal/webhook/v1alpha1/paas_webhook_test.go
devotional-phoenix-97: That would just use default quota's, and the entire quota calculation is in unittests (afaict).
What do you think it would resolve?

internal/webhook/v1alpha2/paas_webhook.go
devotional-phoenix-97: Can but that that would then be in a new module so we can import it in v1alpha1 module and v1alpha2 module.
Is it worth the hazzle for just this? Or should we (maybe) see if we can do this for more duplicate code in webhook and api (but that sounds like a new issue to me)?

internal/webhook/v1alpha2/paasconfig_webhook.go
devotional-phoenix-97: I even built it for v1alpha2 and then copied it to v1alpha1.
Again, we can try to deduplicate, but to me it sounds like something to do on a larger scale.
Like create a module for api, and one for webhook, and try to change duplicate code into functions being called from v1alpha1 or v1alpha2.

By the way, note that we probably need to do stuff with interfaces, and then need to implement methods to be used with the interfaces, and then need to deduplicate those implementations aswell.

If deduplication is something we aim for I personally am okay to work on it in another PR, but don't want to add it in this PR...

internal/webhook/v1alpha2/paasconfig_webhook_test.go
devotional-phoenix-97: No.
- We test when they are not set
- We test when they are all of them set (just to make sure they are all validated
- a subset would do exactly that (validate what is set and not validate what is not set).
We could have another unittest, but there is no worth

internal/webhook/v1alpha1/paas_webhook.go
portly-halicore-76: as `resourceName` is the exactly how the upstream calls the quota spec, I believe we don't derive from that approach then.

internal/webhook/v1alpha1/paas_webhook_test.go
portly-halicore-76: never mind than. Just wanted to make sure nothing breaks.

internal/webhook/v1alpha2/paas_webhook.go
portly-halicore-76: I believe starting now with this approach will be beneficial for the maintenance of the project. Once we start on the next API version and so forth, we reduce the duplication again. I believe we should steer towards this approach for each webhook check, and yes, that can be done in a separate issue, but started now this code is introduced in this PR.

CtrlShiftOps (approved):

PR #751, opened 2025-09-17
fix: Report generator errors back to plugin Generator

portly-halicore-76:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

When an error occurs in generating elements for the plugin generator, the error is swallowed, the Paas entry is skipped by the `continue` keyword. In a situation in which for example the PaasConfig is uninitialized, or a Paas is configured with an not existing cap / customField, the `capElementsFromPaas` func returns an error. This error is caught, however, not passed on to the caller as the result is skipped through the `continue` keyword. Therefore the response to ArgoCD is a 200, without elements for the Paas which generates the error. 

This could, in the worst case, remove a Paas entry entirely from ArgoCD. Imagine the situation in which for the first request from ArgoCD was OK. Entries are returned and the AppSet does it's job. In the meantime, the controller is restarted and experiences a situation in which the PaasConfig is unitialized for a while. The next request from ArgoCD to the plugin generator will in that case return a 200, without any elements, removing all entries from ArgoCD AppSet.

## What is the new behavior?

* Instead of swallowing the error, the error is passed to the caller, resulting in a 500 http response to ArgoCD. ArgoCD does, in that case, not remove an existing entry but mentions the error and continues.
* Updates test cases
* Implements a helper, to reset a PaasConfig (mostly to be used internally in the test)
* Removes outdated comments
* Updates appropriate logging messages

- [ ] Yes
- [x] No

### DIFF
--- a/internal/argocd-plugin-generator/service.go
+++ b/internal/argocd-plugin-generator/service.go
@@ -70,8 +70,8 @@ func (s *Service) Generate(params map[string]interface{}, appSetName string) ([]
 	for _, paas := range paasList.Items {
 		elements, err := capElementsFromPaas(ctx, &paas, capName)
 		if err != nil {
-			logger.Error().Str("paas_name", paas.Name).AnErr("error", err).Msg("failed to get elements")
-			continue // skip failed ones
+			logger.Error().Str("paas_name", paas.Name).AnErr("error", err).Msg("failed to generate elements")
+			return nil, err // return error to caller
 		}
 		if elements == nil {
 			continue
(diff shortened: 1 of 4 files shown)


devotional-phoenix-97:
Hmmm.
[this line](https://github.com/belastingdienst/opr-paas/blob/main/internal/argocd-plugin-generator/service.go#L51C59-L51C69) has a parameter appSetName which is unused, which threw me off.

devotional-phoenix-97 (approved):

PR #757, opened 2025-09-22
feat: Plugin generator as `HasCache` runnable

### LINKED ISSUE #756: feat: Ensure Plugin Generator waits for synced cache via `HasCache`
# Feature Request

Recent PRs show efforts to make the plugin generator more robust and less error-prone. However, one key detail is still missing: implementing it as a HasCache runnable.

The Kubernetes client used in controller-runtime is a split client (see [operator-sdk docs](https://sdk.operatorframework.io/docs/building-operators/golang/references/client/#default-client)). This means that reads such as List are served from the local cache, while writes go directly to the API server. Without proper synchronization, there is a risk that List calls return nothing if the cache has not been filled yet.

By implementing the plugin as a HasCache runnable and wiring informers for both Paas and PaasConfig, we can ensure that the cache is synced during startup before the HTTP server begins serving requests.

* Refactor the plugin generator into a HasCache runnable, with informers for both Paas and PaasConfig.
* Ensure requests are only served once the cache is fully synced.

**Describe Alternatives**

* Use the default Kubernetes client instead of the split client. This would avoid cache sync issues but is not preferred, since the cache is designed to reduce load on the Kubernetes API server.
* Keep the current runnable and add manual sync checks. However, this introduces boilerplate and risks being inconsistently applied.
* Add artificial delays before starting the HTTP server, but this is unreliable and fragile.

**Related Code**

Plugin generator (internal/plugin-generator/*)

**Additional Context**

The absence of cache synchronization can lead to situations where the plugin serves empty or incomplete responses without logging errors. Refactoring into a HasCache runnable will guarantee deterministic startup behavior and improve reliability. The new metric will help operators quickly detect response issues.

_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

appiepollo14:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

A non leader, has no informers added to the managers cache. Therefore, the cache is not synced before retrieving httpRequest to the plugin-generator. This could lead to return no Paas'es, when Listing them, responding with no elements in the Plugin Generator response. 

* Plugin generator is added to the manager as a non-leader runnable
* Plugin generator does not wait for synced caches as non-leader

Fixes: #756 

## What is the new behavior?

The Plugin Generator is added to the manager as a `HasCache` runnable. 
The Plugin Generator adds two informers, `Paas` and `PaasConfig` to the managers cache.
Therefore the Plugin Generator is only started successfully, one the cache is synced.

- [ ] Yes
- [x] No

## Other information

It is really really had to write appropriate unit tests for this. e2e test might by of use, however, to show the strenght of this solution, one should be able to delay the filling of the cache. Which is hard to do IMHO. So if one has any ideas, please respond them.


internal/argocd-plugin-generator/plugin_generator.go
```diff
@@ -62,11 +66,22 @@ func New(kclient client.Client, bindAddr string) *PluginGenerator {
 		TokenEnvVar: tokenEnvVar,
 	}, handler)
 
+	// Ensure informer for Paas exists, to trigger sync of the cache
+	if _, err := c.GetInformer(context.Background(), &v1alpha2.Paas{}); err != nil {
+		return nil, fmt.Errorf("failed to get informer for Paas: %w", err)
```
hikarukin, line 71: I'm assuming here that these errors are being logged higher up and not "eaten"?

hikarukin (changes requested):

internal/argocd-plugin-generator/plugin_generator.go
appiepollo14: Yes. Please see lines 197 till 199 in main.go which is part of this PR

hikarukin (approved):

PR #760, opened 2025-09-23
feat: Optimize `GetConfig` behavior

### LINKED ISSUE #729: feat: Switch internal/config to operator hasCache pattern
# Feature Request

We currently have created a internal/config module, which is loaded as a separate Runnable,
watches PaasConfig changes and updates an internal memory object.
It would be better if we would reimplement this as a hasCache runnable.

That would mean that:

- everywhere in our code, where we use GetConfig, we would instead use a normal Get for a PaasConfig.
- instead of this resulting in a k8s api call, the client would get the latest PaasConfig from the hasCache Runnable
- The hasCache Runnable would have an internal watcher and would update the internally cached version whenever the PaasConfig is updates in k8s

Thus resulting in using the default pattern for fetching k8s resources, without exploding the number of k8s apiece calls, much like what we do now, but in the usual way that the operator-sdk supplies as a default cache pattern.

**Describe Alternatives**

We can stick to what we do now, it is not bad, but it is a custom pattern and has resulted in some unexpected behaviour in the past.

**Related Code**

- https://github.com/belastingdienst/opr-paas/tree/main/internal/config

**Additional Context**

Example implemenation of a structure with the hasCache interface implemented
```
// Filename: has_cache_example.go
// Description: A Go example demonstrating the implementation of a hasCache interface
// for a Kubernetes operator Runnable. This pattern is essential for reducing API server load.

package main

import (
	"context"
	"fmt"
	"sync"
	"time"

	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/runtime/schema"
	"k8s.io/client-go/tools/cache"
)

// HasCache is the interface that a Runnable can implement to expose a shared cache.
// It provides a single method to get the underlying cache.
type HasCache interface {
	GetCache() cache.SharedIndexInformer
}

// PodCacheRunnable is an example of a Runnable that implements the HasCache interface.
// Its purpose is to watch Pod resources and maintain a local, in-memory cache.
type PodCacheRunnable struct {
	// A mock SharedIndexInformer for demonstration purposes. In a real operator,
	// this would be a real informer watching Kubernetes API resources.
	informer cache.SharedIndexInformer
	// A mutex to protect the cache during updates and reads.
	mu sync.RWMutex
}

// NewPodCacheRunnable creates and initializes a new PodCacheRunnable.
func NewPodCacheRunnable() *PodCacheRunnable {
	// For this example, we're creating a dummy informer that will simulate
	// a cache being populated. In a real application, you would use a
	// factory to create a real informer (e.g., corev1.PodInformer()).
	dummyInformer := cache.NewSharedIndexInformer(
		&cache.ListWatch{},
		&runtime.Unstructured{},
		0,
		cache.Indexers{},
	)

	return &PodCacheRunnable{
		informer: dummyInformer,
	}
}

// GetCache implements the HasCache interface. It returns the shared informer,
// which acts as the in-memory cache for other components.
func (p *PodCacheRunnable) GetCache() cache.SharedIndexInformer {
	p.mu.RLock()
	defer p.mu.RUnlock()
	return p.informer
}

// Start is the method required by the controller-runtime's Runnable interface.
// It is responsible for starting the informer and populating the cache.
func (p *PodCacheRunnable) Start(ctx context.Context) error {
	p.mu.Lock()
	defer p.mu.Unlock()

	// In a real scenario, this would start the informer's event loop to watch Pods.
	// For this mock, we'll just simulate populating the cache.
	fmt.Println("PodCacheRunnable: Starting cache population...")

	// Simulate populating the cache with a fake Pod.
	// In a real informer, this is handled automatically.
	dummyPod := &runtime.Unstructured{}
	dummyPod.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Pod"})
	dummyPod.SetName("my-dummy-pod")
	dummyPod.SetNamespace("default")

	// Add the object to the informer's internal store.
	p.informer.GetStore().Add(dummyPod)

	fmt.Println("PodCacheRunnable: Cache populated successfully.")

	// A runnable typically blocks until the context is canceled.
	// In a real application, the informer's `Run` method would do this.
	<-ctx.Done()
	return nil
}

// --- Example of a Controller using the HasCache interface ---

// BackupController is an example of a component that needs Pod data.
// It does not maintain its own cache; instead, it consumes the shared cache.
type BackupController struct {
	podCache HasCache
}

// NewBackupController creates a new controller that receives the HasCache instance.
func NewBackupController(podCache HasCache) *BackupController {
	return &BackupController{
		podCache: podCache,
	}
}

// Reconcile simulates a reconciliation loop where the controller needs to get Pod data.
func (b *BackupController) Reconcile() {
	fmt.Println("BackupController: Starting reconciliation...")

	// Get the shared informer from the HasCache instance.
	informer := b.podCache.GetCache()

	// Use the informer's cache to list all Pods. This is a local operation,
	// not an API call.
	allObjects := informer.GetStore().List()
	fmt.Printf("BackupController: Retrieved %d objects from the shared cache.\n", len(allObjects))
}

// main demonstrates how the components would be wired together in a typical operator setup.
func main() {
	// Create the context for the runnables.
	ctx, cancel := context.WithCancel(context.Background())

	// 1. Create the dedicated runnable responsible for the cache.
	podCacheRunnable := NewPodCacheRunnable()

	// 2. Start the runnable in its own goroutine.
	go func() {
		if err := podCacheRunnable.Start(ctx); err != nil {
			fmt.Printf("PodCacheRunnable failed: %v\n", err)
		}
	}()

	// Give the runnable a moment to start and populate its cache.
	time.Sleep(100 * time.Millisecond)

	// 3. Create a consumer (e.g., a controller) and pass it the HasCache instance.
	backupController := NewBackupController(podCacheRunnable)

	// 4. The controller can now use the shared cache to perform its tasks.
	backupController.Reconcile()

	// Clean up after the example.
	cancel()
}
```

- [x] Yes
- [ ] No

portly-halicore-76:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Custom way to inform on config changes
Multiple funcs to retrieve config etc.

Fixes: #729 

## What is the new behavior?

* Implement `HasCache` interface for Config runnable, to make sure the PaasConfig cache is synced before starting
* Simplify / Unify `GetConfig` methods
* `GetConfig` in an idiomic way with error handling

- [ ] Yes
- [x] No

## Other information

It is a rather large change. I believe it can be further optimised once merged.


internal/webhook/v1alpha1/utils.go
```diff
@@ -83,3 +91,14 @@ func getRsa(ctx context.Context, _c client.Client, paasName string) (*crypt.Cryp
 	crypts[paasName] = c
 	return c, nil
 }
+
+// getConfigFromContext returns the PaasConfig object from the config, using the
+// contextKeyPaasConfig. If the returned value cannot be parsed to a v1alpha1
+// PaasConfig, it returns an error.
+func getConfigFromContext(ctx context.Context) (v1alpha1.PaasConfig, error) {
```
hikarukin: Would this not be better suited to internal/config?

internal/webhook/v1alpha2/utils.go
```diff
@@ -0,0 +1,31 @@
...
+type contextKey int
+
+const (
+	contextKeyPaasConfig contextKey = iota
+)
+
+// getConfigFromContext returns the PaasConfig object from the config, using the
+// contextKeyPaasConfig. If the returned value cannot be parsed to the latest
+// api version PaasConfig, it returns an error.
+func getConfigFromContext(ctx context.Context) (v1alpha2.PaasConfig, error) {
```
hikarukin: Again, the use of "utils" makes me feel this belongs in internal/config? Standard version should return v2, temporary version could return v1?

internal/controller/rsa_controller.go
```diff
@@ -16,8 +15,8 @@ var (
 	decryptPrivateKeys *crypt.PrivateKeys
 )
 
-// resetCrypts removes all crypts and resets decryptSecretPrivateKeys
-func resetCrypts() {
+// ResetCrypts removes all crypts and resets decryptSecretPrivateKeys
+func ResetCrypts() {
```
hikarukin: It feels a little wrong to have a controller export a function like this, but I don't have an alternative right now.

portly-halicore-76: Same feeling here. This is left behind incidentally. It is no longer needed to export this as it is only used in `internal/controller`. Will revert.

internal/webhook/v1alpha1/utils.go
portly-halicore-76: ~~That is a valid option. We need to export it than. Agreable?~~ There are already methods available in that package which retrieve the config from the connected k8s cluster. When adding the config from context methods there, those feel double. The whole point of using the context was to load the config once per reconcilation loop, adding it the that context and retrieve over and over again. WDYT?

portly-halicore-76: Nah we good. Another upside is that we don't always have a k8client to pass to the GetConfig funcs. We do mostly, as is idiomic, have a ctx. That is a main diff which is still valid.

hikarukin (approved):

PR #762, opened 2025-09-29
refactor:  Add funcs on the PaasReconciler struct

appiepollo14:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Some funcs in the controller package, are passed the `Reconciler` as a param. Most others are funcs defined on the reconciler struct.

## What is the new behavior?

All funcs (where applicable), are on the `Reconciler` struct instead of getting the Reconciler passed as an input parameter. This is idiomic Go.

- [ ] Yes
- [x] No

### DIFF
--- a/internal/controller/cluster_wide_quota.go
+++ b/internal/controller/cluster_wide_quota.go
@@ -98,7 +98,8 @@ func (r *PaasReconciler) updateClusterWideQuotaResources(
 }
 
 // backendQuota is a code for Creating Quota
-func backendClusterWideQuota(
+func (r *PaasReconciler) backendClusterWideQuota(
+	ctx context.Context,
 	quotaName string,
 	hardQuotas map[corev1.ResourceName]resourcev1.Quantity,
 ) *quotav1.ClusterResourceQuota {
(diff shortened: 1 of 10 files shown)


internal/controller/cluster_wide_quota.go
```diff
@@ -98,7 +98,8 @@ func (r *PaasReconciler) updateClusterWideQuotaResources(
 }
 
 // backendQuota is a code for Creating Quota
-func backendClusterWideQuota(
+func (r *PaasReconciler) backendClusterWideQuota(
```
devotional-phoenix-97: You change a function into a method, but struct (r) which is not used in the body.
To me a clear sign that it should not be a method.

devotional-phoenix-97 (changes requested):
Check if they should be methods, or can remain functions

internal/controller/cluster_wide_quota.go
appiepollo14: True!

devotional-phoenix-97 (approved):

PR #792, opened 2025-11-11
feat: cap without quota and ns

MidshipGhost:
## added validation so a capability without quota and ns can be made

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

## What is the new behavior?

- when a paasns is created, the quota block of the parent paas can't be empty
- when there is a capability with a paasns, the quota block of the paas can't be removed
- when there is a namespace defined but no quota a namespace won't be created

- [ ] Yes
- [x] No


internal/webhook/v1alpha2/paas_webhook_test.go
```diff
@@ -783,7 +804,70 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+			Expect(causes).To(ContainElements(
+				metav1.StatusCause{
+					Type:    "FieldValueInvalid",
+					Message: "Invalid value: \"1\": quota can not be empty when paas has namespaces (number of namespaces: 1)",
+					Field:   "spec.namespaces",
+				},
+			))
+		})
+
+		It("Should deny creation when a capability with paasNS has been defined without quota", func() {
```
devotional-phoenix-97: Change 'Should deny creation' into 'Should deny modification'.
Change 'defined without quota' into 'modified to have no quota'.

internal/webhook/v1alpha2/paas_webhook_test.go
```diff
@@ -783,7 +804,70 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+			obj = &v1alpha2.Paas{
+				ObjectMeta: metav1.ObjectMeta{Name: paasName},
+				Spec: v1alpha2.PaasSpec{
+					Capabilities: v1alpha2.PaasCapabilities{"cap5": v1alpha2.PaasCapability{}},
+					Quota:        quota.Quota{"limits.cpu": resourcev1.MustParse("10")},
+				},
+			}
+
```
devotional-phoenix-97: Remove this blank line, so you have 3 blocks instead of a lot of separate lines

internal/webhook/v1alpha2/paas_webhook_test.go
```diff
@@ -783,7 +804,70 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+			Expect(k8sClient.Create(ctx, obj)).Error().NotTo(HaveOccurred())
+
+			nsObj := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "my-paas-cap5"}}
+
```
devotional-phoenix-97: Remove this blank line, so you have 3 blocks instead of a lot of separate lines

internal/webhook/v1alpha2/paas_webhook_test.go
```diff
@@ -783,7 +804,70 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+			Expect(k8sClient.Create(ctx, nsObj)).Error().NotTo(HaveOccurred())
+
+			paasNsObj := &v1alpha2.PaasNS{ObjectMeta: metav1.ObjectMeta{Name: "my-paasns", Namespace: "my-paas-cap5"}}
+
```
devotional-phoenix-97: Remove this blank line, so you have 3 blocks instead of a lot of separate lines

internal/webhook/v1alpha2/paas_webhook_test.go
```diff
@@ -783,7 +804,70 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+			Expect(k8sClient.Create(ctx, paasNsObj)).Error().NotTo(HaveOccurred())
+
+			newObj := obj.DeepCopy()
+
```
devotional-phoenix-97: Remove this blank line, so you have 3 blocks instead of a lot of separate lines

internal/webhook/v1alpha2/paas_webhook_test.go
```diff
@@ -783,7 +804,70 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+			newObj.Spec.Quota = nil
+
```
devotional-phoenix-97: Remove this blank line, so you have 3 blocks instead of a lot of separate lines

internal/webhook/v1alpha2/paas_webhook_test.go
```diff
@@ -783,7 +804,70 @@ var _ = Describe("Paas Webhook", Ordered, func() {
...
+			_, err := validator.ValidateUpdate(ctx, obj, newObj)
+
+			var serr *apierrors.StatusError
```
devotional-phoenix-97, line 850: Add a comment to describe this block. Like:
// We want to assert the actual message of the cause, which is packed in the StatusError object type.
// Therefore we type assert, unpack and assert that the Cause is exactly as expected

internal/webhook/v1alpha2/paasns_webhook_test.go
```diff
@@ -311,4 +337,20 @@ var _ = Describe("PaasNS Webhook", Ordered, func() {
 			Expect(warn, err).Error().ToNot(HaveOccurred())
 		})
 	})
+
+	Context("When creating a PaasNs but the parent paas doesn't have a quota defined, ", func() {
+		It("the paasNS can't be created", func() {
```
devotional-phoenix-97: C'ange into 'should not allow creation of a PaasNS'

internal/webhook/v1alpha2/paasns_webhook_test.go
```diff
@@ -311,4 +337,20 @@ var _ = Describe("PaasNS Webhook", Ordered, func() {
...
+			By("creating a paasNs, but the parent paas has no quota defined")
```
devotional-phoenix-97: Change into 'creating a PaasNS for a Paas without a defined Quota block'

internal/controller/cluster_quotas_test.go
```diff
@@ -52,6 +53,9 @@ var _ = Describe("Cluster Quotas", Ordered, func() {
 			},
 			Spec: v1alpha2.PaasSpec{
 				Requestor: paasRequestor,
+				Namespaces: v1alpha2.PaasNamespaces{
```
devotional-phoenix-97, line 58: This seems to fix a case where a quota is defined, and we verify it to be created.
We need a test that a quota is not defined, and that the quota is not created in k8s either.

devotional-phoenix-97 (changes requested):

internal/controller/cluster_quotas.go
```diff
@@ -124,11 +124,32 @@ func (r *PaasReconciler) backendEnabledQuotas(
...
+
+	nsDefs, err := r.nsDefsFromPaas(ctx, paas)
 	if err != nil {
+		logger.Err(err).Msg("could not get nsDefs from paas")
 		return nil, err
 	}
-	quotas = append(quotas, quota)
+	logger.Debug().Msgf("Need to manage resources for %d namespaces", len(nsDefs))
+
+	// if there are paasNs resources or if there are namespaces defined in the paas spec,
```
devotional-phoenix-97: Change this, so that an empty quota block does not end in a quota named after the paas, and a defined block does.
This would be less intrusive when someone has a quotablock defined and creates and deletes a Paas.
This code would then create and delete quota's, the proper implementation would keep the quota if the PaasNs is removed.

devotional-phoenix-97 (changes requested):

devotional-phoenix-97 (approved):

PR #793, opened 2025-11-11
fix: still get go repr values instead of json serialized data in generator output

devotional-phoenix-97:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

- templating module has different types from fields, and other modules use both, so we need to convert
- unnecessary conversion between the types from templater and fields
- fields seems submodule of plugin generator, but is also used by other modules (api, controller, templating, etc.)
- in templating we have MapResult, ListResult and Result, Map and List can convert into Result, no interface, Result is map[string]string

We get json docs with fields set to go representations of complex datatypes

## What is the new behavior?

We should now get json docs with fields  with complex datatypes properly deserialized

- [ ] Yes
- [x] No


internal/fields/elementlist.go
```diff
@@ -0,0 +1,17 @@
+package fields
+
+import (
+	"fmt"
+)
+
+// ElementList represents a list of any values
+type ElementList []Element
+
+// AsElementMap will convert this any map into a string map
```
hikarukin: This comment is off... I would probably rewrite this to something like:

`// AsElementMap converts the list into an ElementMap with string indices as keys.`

internal/fields/elementlist.go
```diff
@@ -0,0 +1,17 @@
...
+func (el ElementList) AsElementMap() (ElementMap, error) {
+	result := ElementMap{}
+	for index, value := range el {
+		result[fmt.Sprintf("%d", index)] = value
+	}
+	return result, nil
```
hikarukin: I don't see the point in returning nil/error since you never return anything other than nil. Just makes for unnecessary error handling.

Go convention is to only return an error when it actually can fail.

internal/fields/elementarray.go
```diff
@@ -0,0 +1,6 @@
+package fields
+
+// ElementArray is an interface which represents all values that could be truned into a string map (a.o.)
```
hikarukin: typo: truned -> turned

Also: I assume "(a.o.)" to mean "amongst others"? Not entirely clear like this, what is meant.

internal/fields/elementlist.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+// ElementList represents a list of any values
+type ElementList []Element
+
+// AsElementMap will convert this any map into a string map
+func (el ElementList) AsElementMap() (ElementMap, error) {
+	result := ElementMap{}
+	for index, value := range el {
+		result[fmt.Sprintf("%d", index)] = value
```
hikarukin, line 14: I assume the conversion to a string index is to support serialization?

internal/fields/elementlist_test.go
```diff
@@ -0,0 +1,36 @@
...
+		element1,
+		element2,
+		element3,
+		element4,
+	}
+)
+
+func TestListAsElementMap(t *testing.T) {
+	sm, convErr := listElements.AsElementMap()
+	assert.NoError(t, convErr)
```
hikarukin: Again, Go convention is to only return an error when something actually can fail.

Since nothing can fail, only nil is ever returned. That makes this unneeded.

Also, your test does not try to test a situation in which an error might be returned, validating the opinion that returning error is unneeded.

internal/fields/elementmap.go
```diff
@@ -0,0 +1,131 @@
...
+func (em ElementMap) AsLabels() map[string]string {
+	result := map[string]string{}
+	for key, value := range em {
+		result[key] = fmt.Sprintf("%v", value)
+	}
+	return result
+}
+
+// AsElementMap will convert this any map into a string map
+func (em ElementMap) AsElementMap() (ElementMap, error) {
```
hikarukin: This feels weird, this looks like a no-op. Also, unneeded error return.

internal/fields/elementmap.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	j, err := json.Marshal(element)
+	if err != nil {
+		return "", err
+	}
+	return string(j), nil
+}
+
+// Merge merges all key/value pairs from another Entries on top of this and returns the resulting total Entries set
+func (em ElementMap) Merge(added ElementMap) ElementMap {
```
hikarukin, line 51: Don't agree with this one... it is confusing/inconsistent.

1. The original `em` is being mutated.
2. The mutated version is returned.
3. Comment talks about Entries, but they're ElementMaps.

I would either:

make it explicit and rename into MergeInPlace returning nothing.

or I'd make it so `em` is immutable, something along the lines of:

```
func (em ElementMap) Merge(added ElementMap) ElementMap {
	result := ElementMap{}
	for k, v := range em {
		result[k] = v
	}
	for k, v := range added {
		result[k] = v
	}
	return result
}
```

internal/fields/elementmap.go
```diff
@@ -0,0 +1,131 @@
...
+func ElementMapFromJSON(raw []byte) (ElementMap, error) {
+	newElementMap := make(ElementMap)
+	if err := json.Unmarshal(raw, &newElementMap); err != nil {
+		return nil, err
+	}
+	return newElementMap, nil
+}
+
+// GetElementMapAsStringMap converts all values to a json value and returns the set as a string map
+func (em ElementMap) GetElementMapAsStringMap() (map[string]string, error) {
```
hikarukin: Strange name. From reading the function I would expect a simple `AsStringMap` or even `ToJSONStrings`?

internal/fields/elementmap.go
```diff
@@ -0,0 +1,131 @@
...
+	keys := make([]string, 0, len(em))
+	for k := range em {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		value := em.GetElementAsString(key)
+		key = strings.ReplaceAll(key, "'", "\\'")
+		value = strings.ReplaceAll(value, "'", "\\'")
+		l = append(l, fmt.Sprintf("'%s': '%s'", key, value))
```
hikarukin: I'm assuming this function is mostly intended for logging purposes, but this does not produce valid JSON is that was the intent?

`l = append(l, fmt.Sprintf("%q: %q", key, value))` is more suitable and would allow you to drop the somewhat confusing strings.ReplaceAll() calls.

internal/fields/elementmap.go
```diff
@@ -0,0 +1,131 @@
...
+	return em
+}
+
+func (em ElementMap) String() string {
+	var l []string
+	keys := make([]string, 0, len(em))
+	for k := range em {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
hikarukin: Is this the reason why the map is returned as a map with string indices? Wouldn't keeping the integer indices auto sort it?

internal/fields/elementmap.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// AsJSON can be used to export all elements as JSON
+func (em ElementMap) AsJSON() ([]byte, error) {
+	return json.Marshal(em)
+}
+
+// Key returns the name of the Paas (as derived from the element with name "paas").
+// ElementMap have key, value pairs, and the "paas" value usually exists and has the name of the Paas.
+func (em ElementMap) Key() string {
```
hikarukin: This is the one function that makes the ElementMap domain specific to a Paas as far as I can tell... perhaps make it more generic?

`func (em ElementMap) GetStringKey(key string) string`

internal/fields/elementmap.go
```diff
@@ -0,0 +1,131 @@
...
+func (em ElementMap) AsElementMap() (ElementMap, error) {
+	return em, nil
+}
+
+// Prefix will return a new ElementMap with all keys prefixed with a value
+func (em ElementMap) Prefix(prefix string) ElementMap {
+	prefixed := ElementMap{}
+	for key, value := range em {
+		if key == "" {
+			key = prefix
```
hikarukin, line 83: This is a little odd and might be unexpected behaviour. I suggest we explicitly mention this behaviour in function level comment so it pops up in the IDE.

internal/fields/entries.go
```diff
@@ -65,7 +65,7 @@ func (en Entries) AsJSON() ([]apiextensionsv1.JSON, error) {
 func EntriesFromJSON(data []apiextensionsv1.JSON) (Entries, error) {
```
hikarukin: My guess is that this is hardwired in K8S? Something like below would make more sense to me:

`func (en Entries) FromJSON(data []apiextensionsv1.JSON) (Entries, error)`

internal/logging/main.go
```diff
@@ -29,6 +30,21 @@ var (
...
 
+// SetPluginLogger derives a context with a `zerolog` logger configured for a plugin request.
+func SetPluginLogger(
+	ctx context.Context,
+	req *http.Request,
+) (context.Context, *zerolog.Logger) {
+	logger := log.With().
+		Str("path", req.URL.Path).
+		Str("method", req.Method).
+		Str("requestID", uuid.NewString()).
```
hikarukin, line 41: This would always generate a new `requestID`. Are we sure we never have a `req` here with an existing `requestID`? If we do, it'd be better (for tracing) to retain the existing ID.

internal/logging/main.go
```diff
@@ -29,6 +30,21 @@ var (
...
 )
 
+// SetPluginLogger derives a context with a `zerolog` logger configured for a plugin request.
+func SetPluginLogger(
+	ctx context.Context,
+	req *http.Request,
+) (context.Context, *zerolog.Logger) {
+	logger := log.With().
+		Str("path", req.URL.Path).
+		Str("method", req.Method).
```
hikarukin, line 40: Do we need to add these?

```
Str("remote_addr", req.RemoteAddr).
Str("user_agent", req.UserAgent()).
```

internal/templating/main.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/belastingdienst/opr-paas/v3/api"
 	"github.com/belastingdienst/opr-paas/v3/api/v1alpha1"
 	"github.com/belastingdienst/opr-paas/v3/api/v1alpha2"
+	"github.com/belastingdienst/opr-paas/v3/internal/fields"
 )
 
 // PaasUnion is an interface representing either a v1alpha1.Paas or a v1alpha2.Paas
```
hikarukin, line 16: Comment doesn't match?

internal/templating/main.go
```diff
@@ -72,18 +73,20 @@ func (t Templater[P, C, S]) TemplateToString(name string, templatedText string)
...
-		return myMap.AsResult(name), nil
+	if myMap, err := yamlToMap([]byte(yamlData)); err == nil {
+		return myMap.Prefix(name), nil
 	}
-	myList, err := yamlToList([]byte(yamlData))
-	if err == nil {
-		return myList.AsResult(name), nil
+	if myList, err := yamlToList([]byte(yamlData)); err == nil {
+		myMap, convErr := myList.AsElementMap()
+		if convErr != nil {
```
hikarukin: AsElementMap() never returns an error, so this is dead code?

internal/templating/main.go
```diff
@@ -72,18 +73,20 @@ func (t Templater[P, C, S]) TemplateToString(name string, templatedText string)
...
-		return nil, err
+func (t Templater[P, C, S]) TemplateToMap(name string, templatedText string) (fields.ElementArray, error) {
+	yamlData, templateErr := t.TemplateToString(name, templatedText)
+	if templateErr != nil {
+		return nil, templateErr
 	}
-	myMap, err := yamlToMap([]byte(yamlData))
-	if err == nil {
-		return myMap.AsResult(name), nil
+	if myMap, err := yamlToMap([]byte(yamlData)); err == nil {
```
hikarukin, line 81: minor nitpick: why convert twice?

```
yamlBytes := []byte(yamlData)
if myMap, err := yamlToMap(yamlBytes); err == nil { ... }
if myList, err := yamlToList(yamlBytes); err == nil { ... }
```

hikarukin (changes requested):

internal/fields/elementlist.go
devotional-phoenix-97: index is int (list index) and ElementMap has string keys instead of indexes.
This is the recommended way to convert and int into a string.

internal/fields/elementlist.go
devotional-phoenix-97: Fixed

internal/fields/elementlist.go
devotional-phoenix-97: Good point. No need indeed. Fixed.

internal/fields/elementmap.go
devotional-phoenix-97: Unnecessary error removed.
This is a No-op, but the only way to implement the ElementArray interface.

Idea was: Function returns an ElementArray (which could be ElementMap or ElementList).
ElementArray can be converted into ElementMap and then further processed.

I checked, and was able to remove the ElementArray entirely, so no-op anymore...

internal/fields/elementmap.go
devotional-phoenix-97: Agreed. I went for the second option (but using maps.clone instead of the first loop).

internal/fields/elementmap.go
devotional-phoenix-97: This method is not even used anymore.
I removed it.

internal/fields/elementlist.go
hikarukin: Check

internal/fields/elementmap.go
devotional-phoenix-97: It was just there for unittest code.
I changed to json.Marshall

internal/fields/elementmap.go
hikarukin: Saw your earlier comment

devotional-phoenix-97: Function is removed.

internal/fields/elementmap.go
devotional-phoenix-97: Yeah, because we don;t have any say in what ArgoCD accepts, we needed to have a map where one key was special (the "paas" key, holding the name of the Paas).
There was 2 functions using this Key method. One was in capabilities. I rewrote it. Much cleaner.
Unfortunately, the other was EntriesFromJSON, which still is in this module.
I rewrote that too, so it expects a key which is used to derive the name, and
in unittesting I set key to "myKey", while in controller logic I set it to "paas".

nice one, thanks.

internal/fields/elementmap.go
devotional-phoenix-97: Added

internal/fields/entries.go
devotional-phoenix-97: ArgoCD list generator structures has json as []byte array here.
We deserialise it to something useful, change it, and serialise it again.
I changed as you mentioned, which is a bit weird, as we create an object just to use the method to deserialise into the actual object. But this is fine.
Note that this code will go once we sunset the ArgoCD list gen code (definitely moved to plugin gen solution).

internal/logging/main.go
devotional-phoenix-97: It should then be in the request. I checked and couldn't find't find anything useful.
Btw, we do this for Controller logger too.

internal/logging/main.go
devotional-phoenix-97: We already did, but in plugin generator code. I moved it here.
We could do it once as part of the first log line, but to me this seems useful.

internal/templating/main.go
devotional-phoenix-97: I don't understand your comment. And This was already there (I think you once added this?).

internal/templating/main.go
devotional-phoenix-97: Was fixed as part of one of your other comments.

internal/templating/main.go
devotional-phoenix-97: You mean yamlToMap and yamlToList, right?
We could now have the yamlToList be part of the yamlToMap code, but I like this as it is more explicit that we try multiple conversions, (one for Map and one for List).

hikarukin: No what I meant is you do `[]byte(yamlDate)` twice (once for yamlToMAp and once for yamlToList) when you could do that only once via yamlBytes.

But like I said: nitpick

internal/fields/elementmap.go
hikarukin: I don't see it?

internal/logging/main.go
hikarukin: OK, if its not there its not there I guess.

internal/templating/main.go
hikarukin: Anyway, I'll resolve it since its a nitpick

hikarukin (approved):

PR #794, opened 2025-11-11
feat: export, restructure, cleanup

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

- We have an exported api that relies on internal codebase
- we have empty / nearly empty modules
- We have a plugin egenerator, which uses an internal struct as an externally available endpoint
- We have come code, which probably can be reused, but is currently internalized
- plugin generator input and output structs are now externally available

## What is the new behavior?

- validate module helt one function, moved to webhook code inline (zero extra code, minus one module)
- fields, groups, quota and templating are now avaiulable als pkg to other projects
- api only relies on extrernally available code

- [ ] Yes
- [x] No

## Other information

- We can use the externally exposed plugin generator structs in our collect-n-merge tool


nexus-core174 (dismissed):

api/paasresource.go
```diff
@@ -1,6 +1,6 @@
 // Package paasresource has a Resource interface which can be used for functions that should work with multiple Paas
```
hikarukin: nitpick: comment still says "Package paasresource..."

internal/controller/paas_controller.go
```diff
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"strings"
 
+	paasapi "github.com/belastingdienst/opr-paas/v3/api"
```
hikarukin, line 15: This piece of the api is not versioned? That feels weird for an api.

hikarukin (changes requested):
One small nitpick and a question.

api/paasresource.go
devotional-phoenix-97: Fixed

hikarukin (approved):

PR #848, opened 2026-01-05
feat: Checking RFC1123 Label Name requirements on Namespace names

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

If a namespace is added to the namespace block, and it's name is not according to RFC-1123 Label naming, the webhook finds it valid, while the controller cannot create a namespace accordingly

## What is the new behavior?

If a namespace is added to the namespace block, and it's name is not according to RFC-1123 Label naming, the webhook finds it invalid

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/internal/webhook/v1alpha1/paas_webhook.go
+++ b/internal/webhook/v1alpha1/paas_webhook.go
@@ -262,11 +262,11 @@ func validatePaasallowedQuotas(
 // RFC 1123 Label Names
 // Some resource types require their names to follow the DNS label standard as defined in RFC 1123.
 // This means the name must:
-// contain at most 63 characters
-// contain only lowercase alphanumeric characters or '-'
-// start with an alphabetic character
-// end with an alphanumeric character
-var rfc1123LabelNamesRegex = regexp.MustCompile(`^[a-zA-Z0-9]([-a-zA-Z0-9]{0,61}[a-zA-Z0-9])?$`)
+// - contain at most 63 characters
+// - contain only lowercase alphanumeric characters or '-'
+// - start with an alphabetic character
+// - end with an alphanumeric character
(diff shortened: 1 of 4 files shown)


internal/webhook/v1alpha1/paas_webhook.go
```diff
@@ -258,6 +259,15 @@ func validatePaasallowedQuotas(
...
 }
 
+// RFC 1123 Label Names
+// Some resource types require their names to follow the DNS label standard as defined in RFC 1123.
+// This means the name must:
+// contain at most 63 characters
+// contain only lowercase alphanumeric characters or '-'
+// start with an alphabetic character
+// end with an alphanumeric character
+var rfc1123LabelNamesRegex = regexp.MustCompile(`^[a-zA-Z0-9]([-a-zA-Z0-9]{0,61}[a-zA-Z0-9])?$`)
```
nexus-core174: The regex does not comply to these rules:

- "Contain only lowercase alphanumeric characters or -": It currently allows uppercase `[a-zA-Z0-9]`
- "Start with an alphabetic character": It allows starting with a digit `^[a-zA-Z0-9]`



```suggestion
var rfc1123LabelNamesRegex = regexp.MustCompile(`^[a-z]([-a-z0-9]{0,61}[a-z0-9])?$`)
```

nexus-core174 (changes requested):

internal/webhook/v1alpha1/paas_webhook.go
devotional-phoenix-97: I agree to the proposed changes, but I checked, required some more changes, and implemented in v1alpha2 too.

hikarukin (approved):

nexus-core174 (approved):

PR #915, opened 2026-03-10
feat: allow specifying an absolute max quota that is allowed during paas submission

hikarukin:
Allow specifying an absolute max quota that is allowed during paas submission by introducing a MaxAllowedSubmittedQuota setting in PaasConfig.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [X] Refactoring (no functional changes, no API changes)
- [X] Build-related changes
- [X] Documentation content changes

Does not exist.

## What is the new behavior?

v1alpha2 only feature.

The `MaxAllowedSubmittedQuota` feature allows administrators to define a global "ceiling" for quotas requested by users in a `Paas` resource.

- [ ] Yes
- [X] No

## Other information

This PR also includes a minor fix for the Makefile, feature docs and a new section in the docs about how to view docs locally.


internal/webhook/v1alpha2/paas_webhook.go
```diff
@@ -628,3 +632,41 @@ func validateAppNamespaceQuota(
...
+		if !ok {
+			// A has a resource B doesn't — treat B's value as zero
+			bQty = resourcev1.MustParse("0")
+		}
+
+		// If A (paas.Spec.Quota) > B (conf.Spec.MaxAllowedSubmittedQuota.MaxQuota)
+		if aQty.Cmp(bQty) > 0 {
+			errs = append(errs, field.Invalid(
+				field.NewPath(pathSpec, "quota"),
+				bQty.String(),
```
nexus-core174: I think this should be `aQty`

internal/webhook/v1alpha2/paas_webhook.go
```diff
@@ -628,3 +632,41 @@ func validateAppNamespaceQuota(
...
+	if len(conf.Spec.MaxAllowedSubmittedQuota.MaxQuota) == 0 {
+		return nil, nil
+	}
+
+	for resource, aQty := range paas.Spec.Quota {
+		bQty, ok := conf.Spec.MaxAllowedSubmittedQuota.MaxQuota[resource]
+		if !ok {
+			// A has a resource B doesn't — treat B's value as zero
+			bQty = resourcev1.MustParse("0")
+		}
```
nexus-core174, line 659: What if an admin only wants to set a limit on say, `requests.cpu`, and no other resources? This would block any Paas that requests those other resources. Not sure if we want this

docs/administrators-guide/max-allowed-submitted-quota.md
nexus-core174, line 1: You should also update the index file to include this new file

nexus-core174 (changes requested):

docs/administrators-guide/max-allowed-submitted-quota.md
hikarukin: I didn't think that made sense since its a configuration option, so I stuck it under configuration.

internal/webhook/v1alpha2/paas_webhook.go
hikarukin: Check, adjusting.

internal/webhook/v1alpha2/paas_webhook.go
hikarukin: Good catch. Altering the setup to skip checks on unset resources.

hikarukin: Adjusted by adding a continue

nexus-core174 (approved):

PR #948, opened 2026-04-08
fix: do not delete default permission CRBs from unrelated capabilities

westerlypeppermint:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

When reconciling, the operator iterates over all capabilities and checks whether a certain default permission is present or not in the paasconfig of that capability and checks for any related CRB. If it encounters a CRB that isn't present in the default permissions config of a capability, it will delete that CRB. However, it also does this for CRBs that are not part of the capability that is currently being reconciled.

## What is the new behavior?

- Before deletion, all CRBs are checked against a list of all default permissions in the current paasconfig. Only if that CRB is not present at all, it will be deleted.

- [ ] Yes
- [x] No

## Other information


internal/controller/clusterrolebindings.go
```diff
@@ -206,39 +206,90 @@ func (r *PaasReconciler) checkForRemovedPermissionsInPaasConfig(
...
+				allDefaultPermissions[role] = sas
+			}
+		}
+	}
+	return allDefaultPermissions, nil
+}
+
+func (r *PaasReconciler) mergeRoles(
+	permissions v1alpha2.ConfigRolesSas,
+	allDefaultPermissions v1alpha2.ConfigRolesSas) map[string]struct{} {
```
devotional-phoenix-97: Cleaner to either merge line 279 and 280 (so that it is a function def split over 2 lines), 
or split into 4 lines (func def, arg1, arg2, return type).

internal/controller/clusterrolebindings.go
```diff
@@ -206,39 +206,90 @@ func (r *PaasReconciler) checkForRemovedPermissionsInPaasConfig(
...
+	rolesToKeep := make(map[string]struct{})
```
devotional-phoenix-97: You could use map[string]bool instead, or you could use slices.Compact.
slices.Compact is cleaner (I should use that more often myself ;) )...

devotional-phoenix-97 (approved):

PR #1074, opened 2026-08-17
feat: go-templating for secrets

### LINKED ISSUE #1056: bug: Secrets are very ssh and ArgoCD specific
# Bug Report

**Paas Operator version:**

v5.9

**Current behaviour:**

You can create secrets at paas level and at capability level, but the actually created secrets are formatted to support ArgoCD specifically.

**Expected behaviour:**

- Secrets should also be usable as tekton secrets
- Secrets should also be usable for other capabilities
- Secrets should be customisable to work with any capability

**Other information:**

I would like to extend the interface so that the format of secrets can be defined with go-formatting.
This format should be a (new) field for a capability.
We could also have it for the entire Paas (when creating secrets outside of a capability.
I would also like to extend the interface so that it would also work with other datatypes (currently only External Secrets come to mind).

devotional-phoenix-97:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Fixes: #1056

## What is the new behavior?

- You can define a go-template that returns a map[string]map[string]string for a capability
- You can define a go-template that returns a map[string]map[string]string for all namespaces
- controller creates secrets as defined by the template
- we have a lot of unittesting in place
- default is as it was
- that is deprecated so that in a future api version default is not to create any secrets
- docs are created

- [ ] Yes
- [x] No


docs/administrators-guide/go-templating.md
```diff
@@ -183,3 +183,120 @@ This ensures that each key is placed on a separate line, and is thus correctly p
     {{end}}{{end}}
     ```
 
+## Secrets with go-templating
+
+The original secret implementation was specifically designed for generating ArgoCD repo secrets, but the secrets where less useful for other capabilities, namespaces, etc.
```
westerlypeppermint: where -> were

docs/administrators-guide/go-templating.md
```diff
@@ -183,3 +183,120 @@ This ensures that each key is placed on a separate line, and is thus correctly p
...
+    This function parses all secrets that apply for this namespace (e.a. capability secrets and paas secrets, or namespace secrets and paas secrets).
+    This option can also be used for PaasConfig defined templates.
+    Next to getPaasSecrets, we have also added getPaasSecret (without the s), this function can be used to get the (unencrpyted) value of a specific secret
+    ( like `{{ mySecret := getPaasSecret "my-secret" }}`).
+
+In the next major version we plan to require a template to be specified, and will then remove this default.
+
+### Adviced template for ArgoCD
+
+Bellow template can be used for the ArgoCD capability:
```
westerlypeppermint: Bellow -> below

docs/administrators-guide/go-templating.md
```diff
@@ -183,3 +183,120 @@ This ensures that each key is placed on a separate line, and is thus correctly p
...
+    {{- end -}}
+    {{- $secrets | toYAML -}}
+    ```
+
+This results in a secret for every paas secret, where the name of the paas secret is used as the url, and the value as the actual secret.
+The name of the secret is derived from the hashed value of the name of the paas secret.
+
+### Adviced template for Tekton
+
+Bellow template can be used for the Tekton capability:
```
westerlypeppermint: same here

docs/administrators-guide/go-templating.md
```diff
@@ -183,3 +183,120 @@ This ensures that each key is placed on a separate line, and is thus correctly p
...
+See [issue 1056](https://github.com/belastingdienst/opr-paas/issues/1056) for more info.
+
+We have now replaced the implementation with a go-templating driven alternative, with allows administrators to define the exact format for secrets per capability, as well as for all namespaces specifically.
+In the future we expect to expand the solution to
+- also support External Secrets integration
+- replace the default template (which currently defaults to the argocd behavior which has very limitted use outside of argocd
+
+### How it works
+
+The PaasConfig has been expanded to allow for .spec.namespace_secrets (for defining a secret template for all namespaces of a Paas), and .spec.capabilities[].secrets (for defing a template for a specific capability).
```
westerlypeppermint: The combination of 'all namespaces of a Paas' and that there are specific options for secrets in capabilities, makes this part slightly unclear to me. I can guess that it works as follows: namespace_secrets only applies to namespaces as defined in spec.namespaces and doesn't influence secrets in namespaces belonging to capabilities, but the docs don't clearly state this at the moment

pkg/templating/main_test.go
```diff
@@ -2,164 +2,191 @@ package templating_test
...
+			}
+		})
+		It("should return err for invalid templates", func() {
+			tpl := templating.NewTemplater(paas, paasConfig)
+			templated, err := tpl.TemplateToMap("invalid", "{{ .NotAPaas.Name }")
+			Expect(err).To(MatchError(ContainSubstring("unexpected \"}\" in operand")))
+			Expect(templated).To(BeEmpty())
+		})
+	})
+	When("using exirafuncs", func() {
```
westerlypeppermint: extraFuncs?

westerlypeppermint (changes requested):

docs/administrators-guide/go-templating.md
devotional-phoenix-97: I rephrased

westerlypeppermint (approved):

